PR #5, opened 2019-12-25
Execute sbt commands in the executor subproject

bastian-koepcke:
This enables building the executor by calling `sbt compile` in the shine root directory.

Depends on [PR #4](https://github.com/rise-lang/executor/pull/4) in the executor repository.


michel-steuwer:
I changed the target, to merge it into `development` (and not into `master`)

PR #24, opened 2020-01-16
[Rise] Change data types, adding some fixed width

Bastacyclop:
https://github.com/rise-lang/rise/pull/15
it seems like there are also a lot of formatting changes :thinking:


bastian-koepcke:
I think this PR shows pretty clearly that the auto formatting doesn't play well with the rise DSL.

PR #27, opened 2020-01-23
Undeclared index identifier

Bastacyclop:
A small example where codegen fails with an undeclared index identifier.
https://github.com/rise-lang/shine/commit/f79a480fe868258b9ad36feaec11d124af4238c0#diff-7630c39c500613b10242256eebeca2bfR20

It seems to me that the imperative DPIA code is correct. However C code generation is bugged:
```c
void foo(int* output, int* e58){
  /* mapSeq */
  for (int i_124 = 0;(i_124 < 12);i_124 = (1 + i_124)) {
    if ((idx125 < 2)) {
      int idx125 = e58[i_124];
      output[i_124] = 0;
    }
     else if ((idx125 < 4)) {
      int idx125 = e58[i_124];
      output[i_124] = 1;
    }
     else {
      int idx125 = e58[i_124];
      output[i_124] = 2;
    } 
  }
}
```

### DIFF
--- a/src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala
+++ b/src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala
@@ -754,10 +754,10 @@ class CodeGenerator(val decls: CodeGenerator.Declarations,
         case C.AST.ArithmeticExpr(ae) => acc(a, env, CIntExpr(ae) :: ps, cont)
         case cExpr:C.AST.Expr =>
           val arithVar = NamedVar(freshName("idxAcc"))
-          acc(a, env, CIntExpr(arithVar) :: ps, generated => C.AST.Block(immutable.Seq(
+          C.AST.Block(immutable.Seq(
             C.AST.DeclStmt(C.AST.VarDecl(arithVar.name, C.AST.Type.int, Some(cExpr))),
-            cont(generated)
-          )))
+            acc(a, env, CIntExpr(arithVar) :: ps, cont)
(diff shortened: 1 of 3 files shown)


Bastacyclop:
I believe the issue comes from:
```scala
    def codeGenIdx(i: Phrase[ExpType],
                   e: Phrase[ExpType],
                   env: Environment,
                   ps: Path,
                   cont: Expr => Stmt): Stmt = {
      exp(i, env, Nil, {
        // [ ... ]
        case cExpr:C.AST.Expr =>
          val arithVar = NamedVar(freshName("idx"))
          exp(e, env, CIntExpr(arithVar) :: ps, generated => C.AST.Block(immutable.Seq(
            C.AST.DeclStmt(C.AST.VarDecl(arithVar.name, C.AST.Type.int, Some(cExpr))),
            cont(generated)
          )))
      })
    }
```
A new variable is created and put on the path, although only the continuation will declare this variable. As a result, the ifs end up using that variable from the path before declaring it.

Bastacyclop:
Generated code after this change:
```c
void foo(int* output, int* e3434){
  /* mapSeq */
  for (int i_3500 = 0;(i_3500 < 12);i_3500 = (1 + i_3500)) {
    {
      int idx3501 = e3434[i_3500];
      if ((idx3501 < 2)) output[i_3500] = 0; else if ((idx3501 < 4)) output[i_3500] = 1; else output[i_3500] = 2;
    } 
  }
}
```

Issue #19, opened 2020-01-14
Unrolling private arrays to scalar variables.


Bastacyclop:
```c
float x37[3]; // private memory
```
would become:
```c
float x37_0;
float x37_1;
float x37_2;
```
It should be possible to perform this unrolling on imperative DPIA.
Combinations of structs and arrays would also be deconstructed to scalar variables in a similar way.

PR #36, opened 2020-02-03
Enforce unique names during substitution

roger-uw:
The unique name invariant enforced before type inference in Rise (see RIse [#19](https://github.com/rise-lang/rise/pull/19)) is preserved during substitution in DPIA.

### DIFF
--- a/src/main/scala/shine/DPIA/Phrases/Phrase.scala
+++ b/src/main/scala/shine/DPIA/Phrases/Phrase.scala
@@ -132,10 +132,51 @@ object Phrase {
   def substitute[T1 <: PhraseType, T2 <: PhraseType](ph: Phrase[T1],
                                                      `for`: Phrase[T1],
                                                      in: Phrase[T2]): Phrase[T2] = {
+    var substCounter = 0
     object Visitor extends VisitAndRebuild.Visitor {
+      def renaming[X <: PhraseType](p: Phrase[X], proc: String => String): Phrase[X] = {
+        case class Renaming(idMap: Map[String, String]) extends VisitAndRebuild.Visitor {
+          override def phrase[T <: PhraseType](p: Phrase[T]): Result[Phrase[T]] = p match {
+            case Identifier(name, t) => Stop(
+              Identifier(idMap(name),
+                VisitAndRebuild.visitPhraseTypeAndRebuild(t, this)).asInstanceOf[Phrase[T]])
+            case Lambda(x, b) =>
+              val newMap = idMap + (x.name -> proc(x.name))
+              Stop(Lambda(VisitAndRebuild(x, Renaming(newMap)).asInstanceOf[Identifier[PhraseType]],
+                VisitAndRebuild(b, Renaming(newMap))))
+            case dl@DepLambda(x, _) =>
+              val newMap = idMap + (x.name -> proc(x.name))
+              Continue(dl, Renaming(newMap))
+            case _ => Continue(p, this)
+          }
+
+          override def nat[N <: Nat](n: N): N = n.visitAndRebuild({
+            case i: NatIdentifier => idMap.get(i.name) match {
+              case Some(newName) => NatIdentifier(newName)
+              case None => i
+            }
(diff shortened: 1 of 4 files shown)


src/main/scala/shine/DPIA/Phrases/Phrase.scala
```diff
@@ -132,10 +132,47 @@ object Phrase {
...
+              val newMap = idMap + (x.name -> freshName(x.name.takeWhile(_.isLetter)))
+              Continue(dl, Renaming(newMap))
+            case _ => Continue(p, this)
+          }
+
+          override def nat[N <: Nat](n: N): N = n.visitAndRebuild({
+            case i: NatIdentifier => idMap.get(i.name) match {
+              case Some(newName) => NatIdentifier(newName)
+              case None => i
+            }
```
Bastacyclop: maybe you could use `map().getOrElse()` instead?

src/main/scala/shine/DPIA/Phrases/Phrase.scala
```diff
@@ -132,10 +132,47 @@ object Phrase {
...
+            case ae => ae
+          }).asInstanceOf[N]
+
+          override def data[T <: DataType](dt: T): T = (dt match {
+            case i: DataTypeIdentifier => idMap.get(i.name) match {
+              case Some(newName) => DataTypeIdentifier(newName)
+              case None => i
+            }
```
Bastacyclop: same

src/main/scala/shine/DPIA/Phrases/VisitAndRebuild.scala
```diff
@@ -49,8 +49,27 @@ object VisitAndRebuild {
           case Apply(p, q) =>
             Apply(apply(p, v), apply(q, v))
 
-          case dl @ DepLambda(a, p) =>
-            DepLambda(a, apply(p, v))(dl.kn)
+          case DepLambda(a, p) => a match {
+            case n: NatIdentifier =>
+              DepLambda[NatKind, PhraseType](
+                NatIdentifier(v.nat(n).asInstanceOf[arithexpr.arithmetic.NamedVar].name), apply(p, v))
```
Bastacyclop: I don't know if there is a better way to solve this issue :thinking:

src/main/scala/shine/DPIA/Phrases/VisitAndRebuild.scala
```diff
@@ -92,14 +111,45 @@ object VisitAndRebuild {
...
+  def visitDataTypeAndRebuild(dataType: DataType, v: Visitor): DataType = dataType match {
+    case i: IndexType => IndexType(v.nat(i.size))
+    case a: ArrayType =>
+      ArrayType(v.nat(a.size), visitDataTypeAndRebuild(a.elemType, v))
+    case vec: VectorType =>
+      VectorType(v.nat(vec.size), v.data(vec.elemType))
+    case r: PairType =>
+      PairType(visitDataTypeAndRebuild(r.fst, v),
+        visitDataTypeAndRebuild(r.snd, v))
+    case _ => v.data(dataType)
```
Bastacyclop: What if `v.data` already deals with the previous cases? then this would ignore it.

src/main/scala/shine/DPIA/Phrases/Phrase.scala
roger-uw: updated

src/main/scala/shine/DPIA/Phrases/Phrase.scala
roger-uw: updated

src/main/scala/shine/DPIA/Phrases/VisitAndRebuild.scala
roger-uw: Updated `visitDataTypeAndRebuild` to use the result from `v.data`, also made it `private` to avoid infinite recursive calls caused by using `visitDataTypeAndRebuild` in `data`.

PR #37, opened 2020-02-04
Faster codegen

Bastacyclop:
The camera pipeline code generation should be 4x faster with the proposed changes (from 4mn to 1mn).

uses https://github.com/elevate-lang/elevate/pull/7


src/main/scala/shine/DPIA/FunctionalPrimitives/Iterate.scala
```diff
@@ -47,7 +48,7 @@ final case class Iterate(n: Nat,
   override def xmlPrinter: Elem = {
     val l = f.t.x
     <iterate n={ToString(n)} m={ToString(m)} k={ToString(k)} dt={ToString(dt)}>
-      <f type={ToString(l ->: ExpType(ArrayType(l, dt), read) ->: ExpType(ArrayType(l /^ n, dt), read))}>
+      <f type={ToString(l ~>: ExpType(ArrayType(l, dt), read) ->: ExpType(ArrayType(l /^ n, dt), read))}>
```
bastian-koepcke: Should this not be a type constructing arrow `->:`?

Bastacyclop: Yeah, it should be. I used intelliJ refactoring for some parts and it messed things up :/
Let me know if you find other mistakes like this.

Bastacyclop: Although ideally we would not need to implement this xmlPrinter manually.

src/main/scala/shine/DPIA/Types/package.scala
```diff
@@ -19,13 +19,13 @@ package object Types {
...
 
   implicit final class ArrowType[T <: PhraseType](t: T) {
-    @inline def ->: (dt: DataType): T = t
-    @inline def ->: (n: Nat): T = t
-    @inline def ->: (addressSpace: AddressSpace): T = t
-    @inline def ->: (accessType: AccessType): T = t
-    @inline def ->: (ntn: NatToNat): T = t
-    @inline def ->: (ntd: NatToData): T = t
-    @inline def ->: (unit: Unit): T = t
+    @inline def ~>: (dt: DataType): T = t
```
michel-steuwer: Could you please explain why you changed the arrow syntax.

src/main/scala/shine/DPIA/Types/DataType.scala
```diff
@@ -46,12 +46,6 @@ final case class DepArrayType private (size: Nat, elemFType: NatToData)
   override def toString: String = s"$size.$elemFType"
 }
 
-object DepArrayType {
-  def apply(size: Nat, f: DPIA.NatIdentifier => DataType): DepArrayType = {
```
michel-steuwer: Was this unused, or why is this removed?

src/main/scala/shine/DPIA/FunctionalPrimitives/Iterate.scala
michel-steuwer: There are quite a few more cases of this ...

src/main/scala/shine/DPIA/Types/package.scala
Bastacyclop: Otherwise there can be ambiguities between `->:` to build a dependent function type, and `~>:` which is only used to specify dependent arguments hidden in fully applied primitives.

Bastacyclop: I also think that this helps avoiding confusion by expressing more intent while reading the code. Are we trying to build a type, or are we just prepending constraint checks?

src/main/scala/shine/DPIA/Types/DataType.scala
Bastacyclop: There is now the following syntax instead:
```
size`.d`f
size`.d`{ x => f(x) }
```

src/main/scala/shine/DPIA/Types/package.scala
Bastacyclop: Additionally, I believe that the refactoring "mistakes" that were introduced such as above in xmlPrinter come from this ambiguity. scala was probably using the wrong arrow before the changes already.

michel-steuwer: I don't think that it's a good idea to introduce new arrow-like syntax - I don't think it is at all clear what this arrow means.

I see that you have removed the `()->` syntax and unified it with the `->:` syntax - which I guess is part of what has made this ambiguous.

I would suggest that we maybe should think about separating the checking of the types of a phrase from the type the phrase has and `t` should return.

So for something like `drop` we would change
```scala
final case class Drop(n: Nat,
                      m: Nat,
                      w: AccessType,
                      dt: DataType,
                      array: Phrase[ExpType]) extends ExpPrimitive {
  override val t: ExpType =
    (n: Nat) ~>: (m: Nat) ~>: (w: AccessType) ~>: (dt: DataType) ~>:
       (array :: expT({n + m}`.`dt, w)) ~>: expT(m`.`dt, w)
// ...
}
```
into something like
```scala
final case class Drop(n: Nat,
                      m: Nat,
                      w: AccessType,
                      dt: DataType,
                      array: Phrase[ExpType]) extends ExpPrimitive {
  array :: expT({n + m}`.`dt, w)

  override val t: ExpType = expT(m`.`dt, w)
// ...
}
```

So essentially every `:` in the current `Phrase.t` values would go and every `::` would become a statement in the class.

Opinions, @bastian-koepcke and @Bastacyclop?

michel-steuwer: This would also emphasize that the primitives in DPIA are not functions, but represent their fully applied counterparts.

Bastacyclop: This looks good to me :+1: I can make the changes if you want.
The only downside is not seeing the full type that we would write on paper (for the unapplied primitive function) anywhere in our code.

michel-steuwer: I think that is fine (and could be changed in the future with the help of macros). The primitives are not functions anyway, so it's a bit confusing to have a type that suggests otherwise anyway.

src/main/scala/shine/DPIA/FunctionalPrimitives/Array.scala
```diff
@@ -14,9 +15,9 @@ final case class Array(dt: DataType,
   extends ExpPrimitive {
 
   private def tRec(m: Int): ExpType =
-    if (m <= 0) { exp"[${elements.length : Nat}.$dt, $read]" }
-    else { dt ->: tRec(m - 1) }
-  override val t: ExpType = (dt: DataType) ->: tRec(elements.length)
+    if (m <= 0) { expT({elements.length : Nat}`.`dt, read) }
+    else { tRec(m - 1) }
```
Bastacyclop: I think that I can entirely remove the recursion and write the final array type `expT(elements.length`.`dt, read)` here.

PR #38, opened 2020-02-06
experimenting with naive strategies for camera pipe

bastianhagedorn:
@bastian-koepcke @Bastacyclop

### DIFF
--- a/src/test/scala/apps/cameraPipeCheck.scala
+++ b/src/test/scala/apps/cameraPipeCheck.scala
@@ -4,12 +4,19 @@ import cameraPipe._
 import util._
 import rise.core.IsClosedForm
 import rise.core.types._
+import rise.core.dotPrinter._
 import rise.core.DSL._
 import rise.core.TypeLevelDSL._
 import elevate.core._
+import elevate.core.strategies.debug.peek
+import elevate.core.strategies.basic._
+import elevate.core.strategies.predicate._
+import elevate.core.strategies.{predicate, basic}
 import elevate.rise.Rise
 import elevate.rise.rules._
-// import elevate.rise.rules.algorithmic._
+import elevate.rise.rules.lowering.{isComputation, mapSeqCompute}
+import elevate.rise.strategies.predicate._
+import elevate.rise.rules.algorithmic._
 // import elevate.rise.rules.movement._
 import elevate.core.strategies.basic._
 import elevate.core.strategies.traversal._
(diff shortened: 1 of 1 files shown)


src/test/scala/apps/cameraPipeCheck.scala
```diff
@@ -105,12 +112,53 @@ float clamp_f32(float v, float l, float h) {
...
+    `try`(
+      // ...there is no computation in the AST...
+      predicate.not(oncetd(isComputation)) `;`
+        // ...insert copy: .
+        oncetd(
+          // ...skip all (Dep-)Lambdas...
+          // ( <+ can be read as `or` ) -> maybe introduce `||` notation?
+          predicate.not(isDepLambda <+ isLambda) `;`
+            // ...insert Id() and transform it to an actual copy
+            idAfter `;` normalize.apply(liftId) `;` oncetd(idToCopy)
```
Bastacyclop: It seems like you could directly use `fun(x => x)` instead of an id primitive here :thinking:

src/test/scala/apps/cameraPipeCheck.scala
```diff
@@ -105,12 +112,53 @@ float clamp_f32(float v, float l, float h) {
...
+        )
+    ) // end-try
+
+  def allIf(cond: Strategy[Rise], s: Strategy[Rise]): Strategy[Rise] =
+    (cond `;` strategies.traversal.all(s)) <+ basic.id[Rise]()
+
+  val p = predicate.not(isApplied(isApplied(isMakeArray)))
```
Bastacyclop: p?

src/test/scala/apps/cameraPipeCheck.scala
```diff
@@ -105,12 +112,53 @@ float clamp_f32(float v, float l, float h) {
...
+  case class traverseWhile(predicate: Strategy[Rise],
+                           s: Strategy[Rise]) extends Strategy[Rise] {
+    def apply(p: Rise): RewriteResult[Rise] =
+      (`try`(s) `;`
+       `if`(predicate)
+          (strategies.traversal.all(traverseWhile(predicate, s))))(p)
+  }
+
+  // todo this is not doing what I want yet!
+  // it should stop traversing as soon as it hits an Apply(makeArray, arg)
```
Bastacyclop: I don't think that looking for `makeArray` should be your goal. I think you should keep track of array dimensions that should be unrolled somehow :thinking: 
Maybe `makeArray` should also be able to not be unrolled .. but then it would probably be sugar for a construct using `generate` and `select`s instead of a separate primitive.

src/test/scala/apps/cameraPipeCheck.scala
```diff
@@ -141,14 +189,10 @@ int main(int argc, char** argv) {
 
   test("deinterleave passes checks") {
     val typed = printTime(infer(nFun(h => nFun(w =>
-      deinterleave(h)(w) >> mapSeq(mapSeq(mapSeq(fun(x => x))))
+      deinterleave(h)(w)  // >> mapSeq(mapSeq(mapSeq(fun(x => x))))
```
Bastacyclop: I believe that elevate is now inserting exactly that, maybe just remove the comment entirely?

src/test/scala/apps/cameraPipeCheck.scala
```diff
@@ -215,10 +259,10 @@ int main(int argc, char** argv) {
...
     val typed = printTime(infer(color_correct))
     println(s"color correction: ${typed.t}")
-    val lower: Strategy[Rise] = LCNF `;` CNF `;`
-      repeatNTimes(2, oncetd(lowering.mapSeq)) `;`
-      oncetd(lowering.mapSeqUnroll)
-    val lowered = printTime(lower(typed).get)
+    // val lower: Strategy[Rise] = LCNF `;` CNF `;`
+    //   repeatNTimes(2, oncetd(lowering.mapSeq)) `;`
+    //   oncetd(lowering.mapSeqUnroll)
+    val lowered = printTime(naiveLowering(typed).get)
```
Bastacyclop: Is this successfully runing without the unrolling? Maybe because of the automatic unrolling we have in the codegen?

Bastacyclop:
Looks promising! But I think we will need some kind of naive fusion decision too: when should we write an intermediate result to memory?

src/test/scala/apps/cameraPipeCheck.scala
Bastacyclop: I think we should also actually talk about "writing to memory" instead of "copying". We make the decision to write an expression in memory, not really to copy it from one memory to another.

michel-steuwer:
This will be submerged by a future PR. I will close this now.

PR #40, opened 2020-02-07
Moved test_util package to avoid name clash

michel-steuwer:
This change avoids confusion :stop_sign: :confused: which `test_util.Test` should be used by which tests.

There are corresponding pull requests in the rise and elevate repos as well.


michel-steuwer:
The failing test is due to the update of the artihexpr dependency, maybe @Bastacyclop has already fixed this in one of the other PRs?

martin-luecke:
Shine does not compile for me anymore with this fix. I get errors in tests which use functions of test_util. 
Example from `shine/src/test/scala/apps/Acoustic3D.scala`
```
Error:(63, 15) object runsWithSameResult is not a member of package test_util
    test_util.runsWithSameResult(Seq(
```

Bastacyclop:
> The failing test is due to the update of the artihexpr dependency, maybe @Bastacyclop has already fixed this in one of the other PRs?

https://github.com/rise-lang/shine/pull/37

PR #44, opened 2020-02-13
Working on camera pipe's performance

Bastacyclop:
This includes:
- better test golds coming from a cropped version of the [example image used by Halide](https://github.com/halide/Halide/blob/master/apps/images/bayer_raw.png)
- in demosaic: interleaving improvement and a small algorithm fix
- the full pipeline now outputs the same image as Halide (tested in another repository)
- manual low-level changes of the Rise expression to get closer to Halide's reference schedule

@michel-steuwer @bastianhagedorn it would be nice if we could now express these first implementation decisions using rewrite rules and elevate.

There is now also a rewrite path from high level demosaic to a low level demosaic implementation using slideSeq (circular buffering).

based on https://github.com/elevate-lang/elevate/pull/14


src/main/scala/shine/DPIA/Compilation/TranslationToImperative.scala
```diff
@@ -131,4 +131,27 @@ object TranslationToImperative {
     }
   }
 
+  def str(E: Phrase[ExpType])
```
bastian-koepcke, line 134: The name is confusing because it seems like it is a function that has to do with `strings`.

Bastacyclop:
included in #57

PR #52, opened 2020-03-26
Harris corner detection (Halide algorithm)

Bastacyclop:
based on https://github.com/rise-lang/shine/pull/44


Bastacyclop:
included in #57

PR #53, opened 2020-05-26
Heuristic search exploration

johanneslenfers:
# Heuristic search exploration
A heuristic search algorithm based exploration tool using elevate strategies. 

#### high priority
- change strategy set  (many rewrites fail or result in invalid expressions in current state)

#### lower priority
- documentation
- command line parser
- merge path together to get full view on exploration
- copy configuration file 
- change hash to hex 
- formatted print to console
- nest heuristics in json configuration file 

### Perform exploration:
execute Test in shine/src/test/scala/exploration/explore.scala (remove @Ignore)

#### Configuration files: 
shine/exploration/configuration/dot_*.json

#### Default output in:
/shine/exploration

Be careful with configurations. Many algorithm nestings, a high depth or high number of iterations may result in very long exploration runs (hours or days).


src/main/scala/exploration/strategies/standardStrategies.scala
```diff
@@ -0,0 +1,192 @@
+package exploration.strategies
```
michel-steuwer, line 1: Is this now redundant with (some of the) strategies that have been added to the shine repo as part of the big merge?

michel-steuwer (approved):

src/main/scala/exploration/strategies/standardStrategies.scala
johanneslenfers: Yes, but they are defined in the tvmGEMM test class

michel-steuwer (approved):

Issue #22, opened 2020-01-14
Simplify code generation pass?


Bastacyclop:
Our current code generation pass is becoming more and more complex. It would be desirable to factor out some logic such as the resolution of indexing primitives (`split`, `take`, ..) to simple `idx` chains.

Bastacyclop:
One related issue is that when unrolling loops for OpenCL, we would like to only deal with loop patterns, but we still have to deal with half-imperative primitives such as `mapRead` that are currently not eliminated before codegen:
https://github.com/rise-lang/shine/blob/da15a58ec6f70095024cff97808d4e0ff45cb34d/src/main/scala/shine/OpenCL/FlagPrivateArrayLoops.scala#L31-L32

Bastacyclop:
An index translation pass is under work in #93

PR #57, opened 2020-09-14
Cam pipe and halide harris merge

Bastacyclop:
Includes #44 and #52.


src/main/scala/shine/C/TranslationContext.scala
```diff
@@ -15,9 +15,12 @@ class TranslationContext() extends shine.DPIA.Compilation.TranslationContext {
...
 
-      //TODO think about this more thoroughly
-      case PairType(_, _) => Assign(dt, lhs, rhs)
-        // assign(dt1, recordAcc1(dt1, dt2, lhs), fst(rhs)) `;` assign(dt2, recordAcc2(dt1, dt2, lhs), snd(rhs))
+      // FIXME: both solutions currently create issues
+      // TODO: think about this more thoroughly
+      // case PairType(_, _) => Assign(dt, lhs, rhs)
+      case PairType(dt1, dt2) =>
+        assign(dt1, pairAcc1(dt1, dt2, lhs), fst(rhs)) `;`
+        assign(dt2, pairAcc2(dt1, dt2, lhs), snd(rhs))
```
Bastacyclop, line 23: I plan to open an issue about this part.

michel-steuwer: yes, please.

src/main/scala/shine/OpenCL/FlagPrivateArrayLoops.scala
```diff
@@ -28,6 +28,9 @@ object FlagPrivateArrayLoops {
       extends VisitAndRebuild.Visitor
     {
       override def phrase[T <: PhraseType](p: Phrase[T]): Result[Phrase[T]] = p match {
+        // FIXME: we would like this primitive to be eliminated before this pass?
```
michel-steuwer, line 31: Can you also create an issue for this one?

src/main/scala/util/package.scala
```diff
@@ -57,4 +65,25 @@ package object util {
...
+  ): Unit = r match {
+    case elevate.core.Success(p) => dotPrintTmp(name, p)
+    case _ =>
+  }
+
+  def dotPrintTmp(name: String, e: rise.core.Expr): Unit = {
+    import scala.language.postfixOps
+    import scala.sys.process._
+
+    val path = "/tmp"
```
michel-steuwer, line 81: Why don't you use the `writeToTempFile` function that uses the `createTempFile` function and avoids hardcoding the `/tmp` path (that e.g. will not work on Windows).

src/test/scala/apps/KMeans.scala
```diff
@@ -8,7 +8,8 @@ class KMeans extends shine.test_util.TestsWithExecutor {
   private val C = 5
   private val F = 34
 
-  test("kmeans versions produce same results") {
+  // FIXME: got broken
+  ignore("kmeans versions produce same results") {
```
michel-steuwer, line 12: Can we create an issue for this as well (and also what does `got broken` mean? does it no longer type check, compiler, pass the test?)

src/test/scala/apps/NBody.scala
```diff
@@ -26,7 +26,8 @@ class NBody extends shine.test_util.TestsWithExecutor {
     ))
   }
 
-  test("nbody AMD version calls update only once") {
+  // FIXME: got broken
+  ignore("nbody AMD version calls update only once") {
```
michel-steuwer, line 30: Same here?

michel-steuwer (approved):
Could you please comment on the individual points I raised?
I am happy with the overall PR and will therefore already approve it, but I still expect your comments.

src/test/scala/apps/KMeans.scala
Bastacyclop: I will write a better comment, but these tests are broken by the pair assignment change.

michel-steuwer: ok. thanks. Please mention them in the issue that you will create.

src/main/scala/shine/C/TranslationContext.scala
Bastacyclop: https://github.com/rise-lang/shine/issues/58 is opened

src/main/scala/shine/OpenCL/FlagPrivateArrayLoops.scala
Bastacyclop: I added a comment in https://github.com/rise-lang/shine/issues/22

src/main/scala/util/package.scala
Bastacyclop: I can use `File.createTempFile` but it should not be deleted on exit, since this is used for debug.

Bastacyclop: Maybe it is better if I provide a path parameter and set `/tmp` as the default?

michel-steuwer: Well than use directly the `File.createTempFile(prefix, suffix)` method from Java.io.File without calling the `deleteOnExit` method.

michel-steuwer (dismissed):

michel-steuwer (approved):

Issue #58, opened 2020-09-15
Pair assignment decision


Bastacyclop:
In the [`TranslationContext`](https://github.com/rise-lang/shine/pull/57/files/da15a58ec6f70095024cff97808d4e0ff45cb34d#diff-d6c5e3324517182e64e6165a34097dad), we can define assignment of pair types in two different ways:
1. a single assignment: `pAcc := pExp`
2. two element-wise assignments: `pAcc.fst := pExp.fst; pAcc.snd := pExp.snd`

We have some examples that are broken because they rely on the first solution, while we currently do the second solution (which useful for other examples):
- https://github.com/rise-lang/shine/blob/da15a58ec6f70095024cff97808d4e0ff45cb34d/src/test/scala/apps/KMeans.scala#L12
- https://github.com/rise-lang/shine/blob/da15a58ec6f70095024cff97808d4e0ff45cb34d/src/test/scala/apps/NBody.scala#L30

fedepiz:
Thinking about this issue, should it be the case that `pAcc := pExp` is always the same as `pAcc.fst := pExp.fst; pAcc.snd := pExp.snd`?
I might touch on the topic soon when I get to assignment of dependent pairs = for which a split-assignment would be very inconvenient to even implement, but should be alright for the `pAcc := pExp` way 
(because snd for dependent pairs it's not really a thing. It's more like `fst` and `fst + snd`)

Bastacyclop:
The concrete issues are roughly:
- with `pAcc := pExp`, we have issues if we start wanting to modify how the inner elements are written (I have this issue for circular buffering as it is currently implemented).
- with `pAcc.fst := pExp.fst; pAcc.snd := pExp.snd` we have issues such as generating `p.fst := f(p).fst; p.snd := f(p).snd` which has the wrong semantics. maybe this should be solved by putting `f(p)` on the stack as a temporary.

bastian-koepcke:
Conceputally, I often go back to grouping our types into to categories when it comes to assignments: *directly assignable* and the rest. Where *directly assignable* talks about C code and whether we can assign a value of the representing C type using a C-assign (`=`). Let's say *directly assignable* types are:
- ScalarTypes/BasicTypes
- Tuples of directly assignable types
(this should model how we can assign in C, where tuples are structs).

Now, for directly assignable types, we could say `pAcc := pExp`; solving the issue that expressions are evaluated twice during tuple assignment.
[...]

PR #62, opened 2020-09-18
Fix tests broken by the element-wise pair assignment

Bastacyclop:
Hacks the continuation translation of foreign functions to avoid calling the function multiple times when triggering a pair assignment.
Removes the acceptor translation of foreign functions, as it should not be needed?

It is open to discussion whether this fix should be merged or if we prefer waiting for a better solution.
Related to #58


src/main/scala/shine/DPIA/FunctionalPrimitives/ForeignFunction.scala
```diff
@@ -34,23 +34,7 @@ final case class ForeignFunction(funDecl: ForeignFunction.Declaration,
...
-          con(arg)(λ(expT(inT, read))(e =>
-            A :=| outT | ForeignFunction(funDecl, inTs :+ inT, outT, exps :+ e)))
-        // with a `tail` of arguments left, recurse
-        case Seq((arg, inT), tail@_*) =>
-          con(arg)(λ(expT(inT, read))(e => recurse(tail, exps :+ e, inTs :+ inT)))
-      }
-    }
-
-    recurse(args zip inTs, Seq(), Seq())
+    ???
```
michel-steuwer, line 37: Can you explain why it is required (or a good idea) to remove the acceptor translation?

src/main/scala/shine/DPIA/FunctionalPrimitives/ForeignFunction.scala
```diff
@@ -64,7 +48,24 @@ final case class ForeignFunction(funDecl: ForeignFunction.Declaration,
         // with only one argument left to process return the assignment of the function call
         case Seq( (arg, inT) ) =>
           con(arg)(λ(expT(inT, read))(e =>
-            C( ForeignFunction(funDecl, inTs :+ inT, outT, exps :+ e) )) )
+            outT match {
+              // TODO: this is an ugly fix to avoid calling the function multiple times
+              //  for pair assignment
```
michel-steuwer, line 53: Could we have an explicit link to the issue discussing this in the comment here?

src/main/scala/shine/DPIA/FunctionalPrimitives/ForeignFunction.scala
Bastacyclop: the `ForeignFunction` primitive is `read`, and so should only trigger continuation translation I believe.

src/main/scala/shine/DPIA/FunctionalPrimitives/ForeignFunction.scala
Bastacyclop: done

michel-steuwer:
:+1:

michel-steuwer (approved):

michel-steuwer:
I suggest that we merge this while keeping the issue open.
It fixes two tests and doesn't seem to introduce harm to the rest of the code base.

src/main/scala/shine/DPIA/FunctionalPrimitives/ForeignFunction.scala
fedepiz: I may be wrong on this, but doesn't a top-level `read` primitive that returns a simple type like `f32` still trigger the acceptor translation?

I think I am facing exactly this in my current implementation of dependent pair match, which in one example is of type `exp[f32, read]`, and triggers the acceptor translation?

Again, I am rusty in `shine`, but it's worth a check

fedepiz (approved):
Looks good

PR #63, opened 2020-09-18
Change the pattern matching style in fromRise

### LINKED ISSUE #61: Change the pattern matching style in fromRise
We should change the pattern matching style in fromRise to avoid breaking unreachablility analysis:
https://github.com/rise-lang/shine/blob/73ea2b4cdc35748e28f87d27f2f9e2e8fb45a18b/src/main/scala/shine/DPIA/fromRise.scala#L106-L120

Instead of matching over the `(p, t)` pair, we could first match over `p` before mathing over `t`. Macros or other helpers could help to refactor this without too much boilerplate changes.

Bastacyclop:
Closes #61

### DIFF
--- a/src/main/scala/shine/DPIA/fromRise.scala
+++ b/src/main/scala/shine/DPIA/fromRise.scala
@@ -1,6 +1,5 @@
 package shine.DPIA
 
-import com.github.ghik.silencer.silent
 import elevate.core.strategies.Traversable
 import elevate.core.strategies.basic.normalize
 import rise.elevate.Rise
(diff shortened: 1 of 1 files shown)


src/main/scala/shine/DPIA/fromRise.scala
```diff
@@ -112,423 +110,543 @@ object fromRise {
     import shine.OpenMP.FunctionalPrimitives._
     import shine.DPIA.Types.MatchingDSL._
 
-    ((p, t): @unchecked) match {
-      case (core.PrintType(msg), expT(dt: DataType, w) ->: _)  =>
+    def fromType(f: PartialFunction[PhraseType, Phrase[_ <: PhraseType]]): Phrase[_ <: PhraseType] = {
```
michel-steuwer, line 113: Nice. I like this.

michel-steuwer (approved):

bastian-koepcke (approved):

PR #64, opened 2020-09-22
Explicit dependence as a togglable flag

fedepiz:
Explicit dependence is necessary for the dependent arrays to work, but makes everything super slow and it seems to critically break arithmetic expression simplification/pivoting in some tests. Therefore, it was removed in the big merge.

I tried to find a general way to have the type-checking only perform explicit dependence when necessary, or to go around the problem entirely. So far, I have been stuck and unable to do so.

Therefore for now, I propose to add a "ExplicitDependence.On/Off" flag, which is passed down to the inference system. It defauls to Off, and it is On only in my dependence tests.

Hopefully infuture we can find a better solution!

### DIFF
--- a/src/main/scala/rise/core/types/Flags.scala
+++ b/src/main/scala/rise/core/types/Flags.scala
@@ -4,7 +4,7 @@ object Flags {
   /** This flag enables the type inference system to express explicit dependence
     * of type variables within dependent functions.
     *
-    * The pass is slow and interferes with advanced arithmetic simplification.
+    * The transformation is slow and interferes with advanced arithmetic simplification.
     * However, it is necessary to use Dependent Arrays properly.
     * */
   sealed abstract class ExplicitDependence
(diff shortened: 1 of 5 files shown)


src/main/scala/rise/core/types/Flags.scala
```diff
@@ -0,0 +1,15 @@
+package rise.core.types
+
+object Flags {
+  /** This flag enables the type inference system to express explicit dependence
+    * of type variables within dependent functions.
+    *
+    * The pass is slow and interferes with advanced arithmetic simplification.
```
Bastacyclop, line 7: it is not really a pass?

Bastacyclop (approved):
Looks fine for short term, is there an issue open to track this?

src/main/scala/rise/core/types/Flags.scala
fedepiz: Good point - will change to "transformation"

fedepiz:
Added an issue, #65, mentioning the details of the problem - and to act as a reminder for the need of a longer-terms solution

Bastacyclop (approved):

michel-steuwer (approved):

PR #66, opened 2020-09-22
Improved @primitive macro

michel-steuwer:
An update to the `@primitive` macro allows the following more concise syntax:

```scala
@primitive object mapSeq extends Primitive with Builder {
    implNat(n => implDT(s => implDT(t =>
      (s ->: t) ->: (n`.`s) ->: (n`.`t))))
}
```
instead of the prior:
```scala
@primitive case class MapSeq()(override val t: Type = TypePlaceholder)
      extends Primitive {
    override def typeScheme: Type = implN(n => implDT(s => implDT(t =>
      (s ->: t) ->: ArrayType(n, s) ->: ArrayType(n, t)
    )))
  }
```

This PR also renames for consistency:
 - `implN` into `implNat`
 - `implT` into `implType`
 - `implA` into `implAddr`
 - `nFunT` into `forallNat`
 - `dtFunT` into `forallDT`
 - `n2nFunT` into `forallN2N`
 - `n2dtFunT` into `forallN2DT`
 - `aFunT` into `forallAddr`
 - `Annotation` into `TypeAnnotation`

### DIFF
--- a/src/main/scala/apps/cameraPipeRewrite.scala
+++ b/src/main/scala/apps/cameraPipeRewrite.scala
@@ -174,7 +174,7 @@ object cameraPipeRewrite {
   def same(toA: Traversal, toB: Traversal): Strategy[Rise] = { p =>
     var a: Rise = null
     toA { e => a = e; Success(e) }(p) flatMapSuccess { p2 =>
-      toB { isEqualTo(erase(a)) }(p2)
+      toB { isEqualTo(eraseTypeFromExpr(a)) }(p2)
     }
   }
 
@@ -347,7 +347,7 @@ object cameraPipeRewrite {
                 // TODO: use proper rewriting to achieve this
                 function(argument(body({ expr =>
                   Success(
-                    typed(expr) |> transpose >> map(transpose) >>
+                    isTyped(expr) |> transpose >> map(transpose) >>
                       // 2 bands of y. all x. rgb channels.
                       mapSeqUnroll(mapSeq(mapSeqUnroll(fun(x => x)))) >>
                       map(transpose) >> transpose
@@ -383,16 +383,16 @@ object cameraPipeRewrite {
 
   def letHoist: Strategy[Rise] = {
     case expr @ App(f, App(App(p.let(), v), Lambda(x, b))) =>
-      Success(letf(lambda(untyped(x), typed(f)(b)))(v) :: expr.t)
+      Success(letf(lambda(eraseType(x), isTyped(f)(b)))(v) :: expr.t)
     // TODO: normal form / non-map specific?
     case expr @ App(App(p.map(), Lambda(y,
       App(App(p.let(), v), Lambda(x, b))
     )), in) if !contains[Rise](y).apply(v) =>
-      Success(letf(lambda(untyped(x), p.map(lambda(untyped(y), b))(in)))(v) :: expr.t)
+      Success(letf(lambda(eraseType(x), p.map(lambda(eraseType(y), b))(in)))(v) :: expr.t)
     case expr @ App(p.map(), Lambda(y,
       App(App(p.let(), v), Lambda(x, b))
     )) if !contains[Rise](y).apply(v) =>
-      Success(fun(in => letf(lambda(untyped(x), p.map(lambda(untyped(y), b))(in)))(v)) :: expr.t)
+      Success(fun(in => letf(lambda(eraseType(x), p.map(lambda(eraseType(y), b))(in)))(v)) :: expr.t)
     case _ => Failure(letHoist)
   }
 
@@ -401,7 +401,7 @@ object cameraPipeRewrite {
     afterTopLevel(
       argument(argument({
         case expr @ App(Lambda(x, color_correct), matrix) =>
-          Success(letf(lambda(toTDSL(x), color_correct))(
+          Success(letf(lambda(toBeTyped(x), color_correct))(
             p.mapSeq(p.mapSeq(fun(x => x)))(matrix)) :: expr.t)
         case _ => Failure(precomputeColorCorrectionMatrix)
       })) `;`
--- a/src/main/scala/apps/cameraPipeRewrite.scala
+++ b/src/main/scala/apps/cameraPipeRewrite.scala
@@ -1,7 +1,7 @@
 package apps
 
 import util.printTime
-import rise.core._
+import rise.core.{primitives => p, _}
 import rise.core.TypedDSL._
 import elevate.core._
 import elevate.core.strategies.basic._
@@ -71,12 +71,12 @@ object cameraPipeRewrite {
     if (predicate(p)) { a(p) } else { b(p) }
   }
 
-  def isAppliedMap: Strategy[Rise] = function(function(isEqualTo(DSL.map)))
-  def isAppliedZip: Strategy[Rise] = function(function(isEqualTo(DSL.zip)))
-  def isAppliedDrop: Strategy[Rise] = function(depFunction(isEqualTo(DSL.drop)))
-  def isAppliedTake: Strategy[Rise] = function(depFunction(isEqualTo(DSL.take)))
+  def isAppliedMap: Strategy[Rise] = function(function(isEqualTo(p.map.primitive)))
+  def isAppliedZip: Strategy[Rise] = function(function(isEqualTo(p.zip.primitive)))
+  def isAppliedDrop: Strategy[Rise] = function(depFunction(isEqualTo(p.drop.primitive)))
+  def isAppliedTake: Strategy[Rise] = function(depFunction(isEqualTo(p.take.primitive)))
   def isAppliedSlide: Strategy[Rise] =
-    function(depFunction(depFunction(isEqualTo(DSL.slide))))
+    function(depFunction(depFunction(isEqualTo(p.slide.primitive))))
 
   def anyMapOutsideZip: Strategy[Rise] = {
     // function(function(isEqualTo(DSL.zip))) `;`
@@ -156,8 +156,8 @@ object cameraPipeRewrite {
 
   // idx i >> f -> map f >> idx i
   def idxAfterF: Strategy[Rise] = {
-    case expr @ App(f, App(App(primitives.Idx(), i), in)) =>
-      Success(idx(i, map(f, in)) :: expr.t)
+    case expr @ App(f, App(App(p.idx(), i), in)) =>
+      Success(p.idx(i)(p.map(f)(in)) :: expr.t)
     case _ => Failure(idxAfterF)
   }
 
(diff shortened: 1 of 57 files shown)


src/main/scala/rise/core/Builder.scala
```diff
@@ -0,0 +1,22 @@
...
+    TypedDSL.app(TypedDSL.toTDSL(apply), e)
+  def apply(n: Nat): TypedDSL.TDSL[DepApp[NatKind]] =
+    TypedDSL.depApp[NatKind](TypedDSL.toTDSL(apply), n)
+  def apply(dt: DataType): TypedDSL.TDSL[DepApp[DataKind]] =
+    TypedDSL.depApp[DataKind](TypedDSL.toTDSL(apply), dt)
+  def apply(a: AddressSpace): TypedDSL.TDSL[DepApp[AddressSpaceKind]] =
+    TypedDSL.depApp[AddressSpaceKind](TypedDSL.toTDSL(apply), a)
+
+  def unapply(arg: Expr): Boolean =
+    throw new Exception("unapply method must be overridden")
```
Bastacyclop, line 18: I find it confusing that `unapply` deconstructs an `Expr` while `apply` constructs a `TDSL[Expr]`, especially with the additional `primitive` function on the side.

src/main/scala/rise/core/package.scala
```diff
@@ -40,4 +42,8 @@ package object core {
       case p: Primitive        => p.toString
     }
   }
+
+  implicit def primitiveBuilderToPrimitive(pb: Builder
```
Bastacyclop, line 46: maybe `Builder` should be `PrimitiveBuilder`?

Bastacyclop (commented):
The macro syntax is very nice, but I think that the different primitive usages (pattern matching, Expr DSL and TDSL) are more confusing than before.

`P extends Primitive with Builder` but we have to use `P.primitive : Primitive`, should there be only `P: PrimitiveBuilder` ?

Otherwise what was the consistency issue with `nFunT` versus `forallNat`? In fact, I expected `forallNat` to be more representative of `implNat`, especially in the primitive type schemes. But maybe this is just personal bias.

michel-steuwer:
Thanks for your comments @Bastacyclop.

I agree that the current mixture of `DSL`, `TDSL`, and the pattern matching is confusing.
I would like to eventually remove the untyped `DSL` and evolve the `TDSL` to be the only _DSL_ to use.

I did not want to combine both of these changes into a single PR.
Therefore, some changes involving the `DSL` are not how I would like it to be eventually.

src/main/scala/apps/cameraPipeRewrite.scala
```diff
@@ -71,12 +71,12 @@ object cameraPipeRewrite {
     if (predicate(p)) { a(p) } else { b(p) }
   }
 
-  def isAppliedMap: Strategy[Rise] = function(function(isEqualTo(DSL.map)))
-  def isAppliedZip: Strategy[Rise] = function(function(isEqualTo(DSL.zip)))
-  def isAppliedDrop: Strategy[Rise] = function(depFunction(isEqualTo(DSL.drop)))
-  def isAppliedTake: Strategy[Rise] = function(depFunction(isEqualTo(DSL.take)))
+  def isAppliedMap: Strategy[Rise] = function(function(isEqualTo(p.map.primitive)))
```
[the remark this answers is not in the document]
michel-steuwer: These could have remained with the `DSL.primitive` syntax.
As explained in another comment, I would like to see that we have only a single _DSL_ eventually.
This is one possibility of how this could then looks like (maybe a conversion from the TDSL type to the primitive could also be done implicitly, but I am not yet sure if that would be a good idea).

src/main/scala/rise/elevate/rules/algorithmic.scala
```diff
@@ -169,46 +169,46 @@ object algorithmic {
...
   }
 
   @rule def mapIdentity: Strategy[Rise] = {
-    case expr @ App(Map(), Lambda(x1, x2)) if x1 == x2 => Success(fun(x => x) :: expr.t)
+    case expr @ App(map(), Lambda(x1, x2)) if x1 == x2 => Success(fun(x => x) :: expr.t)
   }
 
   // x -> join (slide 1 1 x)
-  @rule def slideAfter: Strategy[Rise] = e => Success(join(slide(1)(1)(e)) :: e.t)
+  @rule def slideAfter: Strategy[Rise] = e => Success(join(slide(1: Nat)(1: Nat)(e)) :: e.t)
```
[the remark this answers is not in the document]
michel-steuwer: There is an implicit conversion (`Int => Nat`) that is not fired anymore.
Alternatively, we could write `Cst(1)` or make an additional overload to the `Builder` (and possibly the `TDSL`).

src/main/scala/rise/elevate/rules/movement.scala
```diff
@@ -384,17 +384,17 @@ object movement {
 
         case _ =>
           // todo implement recursively
-          val reduceArgTransposed = inputT match {
+          val reduceArgTransposed: TDSL[Rise] = inputT match {
             case ArrayType(_, ArrayType(_, ArrayType(_,ArrayType(_,_)))) =>
-              transpose o map(transpose) o map(map(transpose))
+              transpose.apply >> map(transpose.apply) >> map(map(transpose))
```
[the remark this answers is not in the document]
michel-steuwer: Yep. I agree. I will fix this.

src/main/scala/rise/core/Builder.scala
michel-steuwer: The `TDSL` performs type inference on construction using the `TDSL` auxiliary type.
This is what is constructed with the `apply` method.
The `unapply` method is used for pattern matching in the rewriting where the `TDSL` type has been converted into the standard `Expr` type (after type inference).

The design of the `TDSL` is inherently asymmetrical: it uses the `TDSL` type for making the type inference work, but afterward (when you want to match) this type will be gone again.
The `Builder` trait basically reflects this design.

src/main/scala/rise/elevate/rules/algorithmic.scala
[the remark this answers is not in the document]
Bastacyclop: Writing `: Nat` or re-introducing the implicit conversion are both fine with me. I think `Cst(1)` is not as nice.

src/main/scala/rise/core/package.scala
michel-steuwer: That was the name I had before. I also thought of making it `Primitive.Builder`.
Opinions?

src/main/scala/apps/cameraPipeRewrite.scala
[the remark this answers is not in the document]
Bastacyclop: Maybe we could use something like `untyped(map)` to avoid triggering type inference when not desired (I think type inference is triggered on TDSL -> Expr conversion otherwise, but this would bypass it, I need to double check the code).

Bastacyclop:
Instead of having `Expr` and `TDSL[Expr]`, maybe we could have a clearer distinction between untyped expressions and typed expressions. Then the DSL builds untyped expressions and type inference is triggered when a typed expression is needed (excatly like TDSL I believe).

michel-steuwer:
You haven't confinced me of the beenfit of this.
Isn't this just shifting the more complex type from one end to the other?

The `TDSL` was really helpful in the rewriting where we want to preserve the typing. I see this as a big plus.
In my experience the `TDSL` is working well. But maybe there are other experiences out there?
Do you have concrete examples that are akward in the `TDSL` (and that we can't resolve easily)?
Additional experiences @fedepiz and @bastian-koepcke?

The main problem I currently see is that there are two _DSLs_ with different style and that obviously adds confusion.

Bastacyclop:
I think the benefit is mainly to avoid confusion, it would not change the underlying mechanisms. But it would be clear when an expression is typed or untyped. Currently `Expr` can be both typed or untyped, while `TDSL[Expr]` is always untyped. I suggest that `Expr` should always be untyped (it then becomes equivalent to TDSL), while `Typed[Expr]` would always be typed (currently has no equivalent).

michel-steuwer:
After our discussion today at the meeting, I have renamed `TDSL` into `ToBeTyped`.

src/main/scala/rise/core/TypedDSL.scala
```diff
@@ -279,23 +279,23 @@ object TypedDSL {
...
 
-  final case class TDSL[+T <: Expr](private val e: T) {
+  final case class ToBeTyped[+T <: Expr](private val e: T) {
     def toExpr: Expr = TDSL.infer(e)
-    def >>=[X <: Expr](f: T => TDSL[X]): TDSL[X] = f(e)
+    def >>=[X <: Expr](f: T => ToBeTyped[X]): ToBeTyped[X] = f(e)
   }
 
-  implicit def typed[T <: Expr](e: T): TDSL[Opaque] = TDSL(Opaque(e)())
+  implicit def isTyped[T <: Expr](e: T): ToBeTyped[Opaque] = ToBeTyped(Opaque(e)())
```
Bastacyclop, line 287: or maybe `keepType`/`preserveType`? otherwise it sounds like a predicate.

Bastacyclop (approved):
LGTM, but I don't think that you answered these previous comments:
> P extends Primitive with Builder but we have to use P.primitive : Primitive, should there be only P: PrimitiveBuilder ?
>
> Otherwise what was the consistency issue with nFunT versus forallNat? In fact, I expected forallNat to be more representative of implNat, especially in the primitive type schemes.

src/main/scala/rise/core/TypedDSL.scala
michel-steuwer: Yes, I like that. I will choose `preserveType`.

michel-steuwer:
Answering the two questions raised earlier:

> P extends Primitive with Builder but we have to use P.primitive : Primitive, should there be only P: PrimitiveBuilder ?

tl;dr: It's a bit of a hack to avoid having the IDE complain about pattern matching for some primitives.

In the macro implementation there are two different cases to consider:
Simple primitives without meta-data, where the `Builder` is an `object`, and primitives with meta-data where the `Builder` is a `case class`.
For pattern matching the `Builder` contains the `unapply` method that will be implemented for the simple patterns in their `Builder` `objects`.
But the `Builder` trait does not contain the `unapply` allowing you to match the meta-data.
The macro does generate a case-class that will allow for this, but as the IDE will not see the generated classes it gets confused and complains that these matches are pointless.
By making the `@primitve` annotated declarations also inheret from `Primitive` the IDE will not produce these warnings and eventually resolve to the correctly generated method.


> Otherwise what was the consistency issue with nFunT versus forallNat? In fact, I expected forallNat to be more representative of implNat, especially in the primitive type schemes.

The comment on the consistency was to point out the `implN` (now `implNat`) had the `n` last where the `nFunT` had it come first.
In addition, `nFunT` doesn't really mean anything. Where `forallNat` is much more meaningful.
I am open for suggestions on a better naming scheme that includes `implNat` and `forallNat` (I personally rejected `implForallNat` and similar too long ideas).

Bastacyclop (approved):

fedepiz (approved):
Looks good to me

PR #73, opened 2020-10-11
Initial Shine support for Dependent Pair & Dependent Arrays

fedepiz:
This PR implements basic support (primitives and code generation) for Dependent Pairs, and also re-introduces support for the Dependent Arrays.
It's initial, in the sense that performance has not been considered yet - particularly, not materialising DepPair constructions/destruction pairs.


src/main/scala/rise/core/primitives.scala
```diff
@@ -169,7 +169,7 @@ object primitives {
   }
 
   @primitive object dpair extends Primitive with Builder {
-    impl{ fdt: NatToData => expl((n: Nat) =>
+      impl{ fdt: NatToData => expl((n: Nat) =>
```
Bastacyclop, line 172: why change the indentation?

src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala
```diff
@@ -163,6 +166,35 @@ class CodeGenerator(val decls: CodeGenerator.Declarations,
...
+            C.AST.DeclStmt(
+              C.AST.VarDecl(fstId.name, C.AST.Type.u32, Some(C.AST.ArraySubscript(C.AST.Cast(C.AST.PointerType(C.AST.Type.u32), input), C.AST.Literal("0"))))
+            ),
+            C.AST.DeclStmt(
+              C.AST.VarDecl(sndId.name, sndCType,
+                Some(initT(C.AST.Cast(sndCType, C.AST.BinaryExpr(input, C.AST.BinaryOperator.+, C.AST.Literal("4")))))
+            )),
+            cmd(f(fstId)(sndId), env.updatedIdentEnv((sndId, C.AST.DeclRef(sndId.name))))
+          ))
+        })
```
Bastacyclop, line 190: would it be better/possible to use a C struct instead of casts? (potentially re-using regular tuple codegen)

fedepiz: I don't think so. A nat-dependent pair with an array element is pretty much a dynamic-sized array whose length is encoded as a payload in the first element.
For now, the strategy is I am going for it represent the type `(Nat ** (n => T(n)))` (dependent pair where the first element is a nat and the second element some data-type T which may depend on n), is to represent it as a single `uint8_ t* buffer`, from which the first element is projected out at offset 0 (by doing `uint32_t first = ((uint32_t*)buffer)[0]`, and then use that identifier to know the size of the second element.

A struct doesn't work, because we'd need conceptually something like
``` struct {  T1 n; T2 something[E(n)] }```. where E is some arithmetic expression.

Let me know if this makes sense, or I have misunderstood you along the way

(PS - originally, I wanted to represent them with structs of pointers to data. This would allow an easier access, but that would mean that a dependent  pair is a struct + two buffers, and that makes GPU-CPU interaction hard, and also  adds further complications when it comes to allocating these things. So ultimately, I believe the "one (materialised) pair is one buffer" approach works best)

File: src/main/scala/rise/core/primitives.scala (hunk 1 above)
Comment (fedepiz) [reply src/main/scala/rise/core/primitives.scala:172]: Merging master weirdness. Will fix

File: src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala (hunk 2 above)
Comment (Bastacyclop) [reply src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala:190]: Would the struct not look more like `struct { T1 n; T2* array; }`, where the size of the array is erased in C? Although this is only for arrays, maybe it does not work for the general case. You could also have `struct { T1 n, uint8_t* buffer }`, although I am not sure if that would bring much benefit. Maybe you could use a combination of tuples and casts in imperative DPIA to express the dependent pair logic without changing the lowest level codegen.

Comment (fedepiz) [reply src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala:190]: With that struct approach, I lose the property that 1 dependent pair = 1 buffer, no? The second element would be a pointer, not an actually buffer itself. That is quite a radical departure from our model, where 1 "concrete" value is actually 1 "contiguous" buffer.

About elimination of dependent patterns before codegen, I see the value of it, but I am not sure how to actually do it and preserving  the properties I want.

That being said, it might be worth it to perhaps  push some of the complexity out of codegen and into the imperative language, I guess. This I think may be done. 

About turning things into pairs, I think this won't be possible as long as records (materialised pairs) are represented as C structs. I believe our code generator system already falters if you try to generate a record of arrays. This I think has been fine so far as we never wanted to do such a thing - but with dependent pairs, that's kind of the whole point

Comment (Bastacyclop) [reply src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala:190]: Sounds reasonable. I guess it would be easier to start by pushing things like split/join out of the monolithic codegen rather than this.

Comment (fedepiz) [reply src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala:190]: Actually about that, I wonder how much do you think should we push out? I can think it's  at least two different "levels" of work here: one about turning `split/join` into multi-dimensional indexing, and a further level of index flattening.  Would you ideally like both to be done as transforms over imperative `DPIA`?

Comment (Bastacyclop) [reply src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala:190]: It would be about turning `split/join` to multi-dimensional indexing, breaking it down to `idx`. The flattening can remain in the `DPIA -> C` part.

Comment (fedepiz) [reply src/main/scala/shine/C/CodeGeneration/CodeGenerator.scala:190]: Makes sense

File: src/main/scala/shine/OpenCL/Kernel.scala [hunk 3]
```diff
@@ -367,6 +368,7 @@ case class Kernel(decls: Seq[C.AST.Decl],
         case _: NatToDataIdentifier =>
           throw new Exception("This should not happen")
       }
+    case _: DepPairType => ???
```
michel-steuwer, line 371: Can you make the `???` an explicit exception as in the lines below.

michel-steuwer (commented):
I am happy with the PR if you could fix the small issues raised by myself and  Thomas.

michel-steuwer (approved):

PR #74, opened 2020-10-12
Tiled Harris operator with prototype uneven tiling support

Bastacyclop:
Investigates multiple options to support uneven tiling in Rise (array is not dividable by the tile size).
- Prototype shift inwards implemention inspired from the [Halide tail strategy](https://halide-lang.org/docs/namespace_halide.html#a6c6557df562bd7850664e70fdb8fea0f). The current implementation seems to work on the tested hardware, but following the strict OpenCL model it creates a race condition. A scatter pattern which goes against the non-overlapping rule on array acceptor elements is used and creates this race condition. Halide creates exactly the same race condition, which it considers as benign. Alternatively to get rid of the race condition, we could consider using atomic stores.
- Prototype epilogue implementation. Currently it does not type check due to nat unification issues. Additionally it uses two distinct `mapGlobal`s that will not spread the thread range as desired.
- Prototype dependent array implementation. This implementation is promising but is also difficult to type check.


Bastacyclop:
uses https://github.com/rise-lang/arithexpr/pull/12

Bastacyclop:
uses https://github.com/rise-lang/arithexpr/pull/13

src/test/scala/apps/harrisCornerDetectionHalideCheck.scala
```diff
@@ -184,6 +187,26 @@ class harrisCornerDetectionHalideCheck
...
+    import rise.openCL.TypedDSL._
+    import rise.core.primitives.mapSeq
+
+    checkOCL(lowerOCL(
+      ocl.harrisTileShiftInwardsPar(tileX, tileY, mapGlobal(_),
+      ocl.harrisVecUnaligned2(4, _ => mapSeq, toPrivate))),
+      LocalSize((1, 1)), GlobalSize((32, 32)))
+  }
+
+  // FIXME: does not pass in CI
```
Bastacyclop, line 200: I will remove that comment

src/test/scala/rise/elevate/movement.scala
```diff
@@ -21,8 +21,10 @@ class movement extends test_util.Tests {
   def betaEtaEquals(a: Rise, b: Rise): Boolean = {
     val na = BENF(a).get
     val nb = BENF(b).get
-    val uab: Rise = toBeTyped(na) :: nb.t
-    makeClosed(uab) == makeClosed(nb)
+    val uab: Rise = eraseType(na) :: nb.t
```
Bastacyclop, line 24: I am not sure why erasing types is necessary .. but otherwise this test randomly fails.

michel-steuwer: Could you create an issue with a link to this so that we can investigate this?

michel-steuwer (dismissed):

src/test/scala/rise/elevate/movement.scala
Bastacyclop: https://github.com/rise-lang/shine/issues/86

michel-steuwer (approved):

PR #82, opened 2020-12-07
More image processing applications

Bastacyclop:
This PR adds the following applications with output checks on a naive low-level Rise program:
- unsharpMask which sharpens an image
- multiscaleInterpolation which can be used for anti-aliasing
- localLaplacian which improves contrast


michel-steuwer (changes requested):
I am fine with all code changes.

Could we introduce a top-level folder `data`, `assets` or something like this to move the `golds` folder in. This could then also hold sample images or other data that is used in the code.

Bastacyclop:
@michel-steuwer I introduced a `data` folder.

michel-steuwer (approved):
Great. Thanks!

PR #83, opened 2020-12-07
Slow code generation issue with code size blowup

Bastacyclop:
This example expression takes around 4mn to compile and generates huge code (about 60 million characters) with too many ifs.


michel-steuwer (approved):
👍

PR #88, opened 2021-01-08
Allow clang to take unused cli arguments given by NixOS's clang wrapper

umazalakain:
Otherwise -Werror makes errors out of these warnings.


umazalakain:
Somehow this is not an issue anymore (checked on a freshly checked out shine repo)

PR #92, opened 2021-01-21
Host code generation

Bastacyclop:
This PR contains work in progress for basic host code generation.


Bastacyclop:
The implementation is currently very ugly and hacked together, but here is the current generated code for one example:
```scala

  test("basic kernel call with variable size and post-process") {
    val e = depFun((n: Nat) => fun((n`.`i32) ->: (n`.`i32))(in =>
      oclRun(LocalSize(2), GlobalSize(n/2))(mapGlobal(add(li32(1)))(in)) |> toMem |> mapSeq(mul(li32(2)))
    ))
    dumpModule(gen.opencl.hosted.fromExpr(e))
  }
```
```c
__kernel __attribute__ ((reqd_work_group_size(2, 1, 1)))
void k0(global int* restrict output, int n1, const global int* restrict e2){
  /* Start of moved local vars */
  /* End of moved local vars */
  /* mapGlobal */
  for (int gl_id_103 = get_global_id(0);(gl_id_103 < n1);gl_id_103 = (gl_id_103 + (n1 / 2))) {
    output[gl_id_103] = (((int)1) + e2[gl_id_103]);
  } 
}
```
```c
#include "runtime.h"


void foo(Context ctx, Buffer moutput, int n1, Buffer me2){
  {
    Buffer mx81 = createBuffer(ctx, (4*n1), 0/* TODO */);
    {
      Kernel k0 = loadKernel(ctx, "k0", "k0.cl");
      cl_mem b0 = targetBufferSync(ctx, mx81, (4*n1), TARGET_WRITE);
      cl_mem b2 = targetBufferSync(ctx, me2, (4*n1), TARGET_READ);
      clSetKernelArg(k0, 0, sizeof(cl_mem), (&(b0)));
      clSetKernelArg(k0, 1, 4, (&(n1)));
      clSetKernelArg(k0, 2, sizeof(cl_mem), (&(b2)));
      const size_t global_size[3] = ((size_t[3]){(n1 / 2), 1, 1});
      const size_t local_size[3] = ((size_t[3]){2, 1, 1});
      launchKernel(ctx, k0, global_size, local_size);
      destroyKernel(ctx, k0);
    }
    {
      int32_t* x81 = ((int32_t*)hostBufferSync(ctx, mx81, (4*n1), HOST_READ));
      int32_t* output = ((int32_t*)hostBufferSync(ctx, moutput, (4*n1), HOST_WRITE));
      /* mapSeq */
      for (int i_84 = 0;(i_84 < n1);i_84 = (1 + i_84)) {
        moutput[i_84] = (((int32_t)2) * x81[i_84]);
      }
    }
    destroyBuffer(ctx, mx81);
  }
}
```

Bastacyclop:
I put the runtime C implementation in `data/runtime` but I think it is not a great location, should it be in `src/runtime` maybe, or even just `./runtime`? @michel-steuwer

michel-steuwer:
> I put the runtime C implementation in `data/runtime` but I think it is not a great location, should it be in `src/runtime` maybe, or even just `./runtime`? @michel-steuwer

I think `/runtime` is good. I would avoid both `/data` and `/src` as they should be reserved for data assets and Scala / Java source code instead.

src/main/scala/util/gen.scala
```diff
@@ -2,56 +2,88 @@ package util
...
+                                          shine.C.CodeGenerator()
+                                      ): Expr => String =
+    functionFromExpr(name, gen) andThen
+      C.Module.translateToString andThen
+      run(SyntaxChecker(_))
+
+  private def functionAsString: CModule => String =
+    // FIXME: SyntaxChecker disabled for host code prototype
+    C.Module.translateToString /*_ andThen
+      run(SyntaxChecker(_))*/
```
Bastacyclop, line 68: I think that including the runtime header during syntax checking would solve this one.

src/main/scala/util/gen.scala
```diff
@@ -123,54 +155,54 @@ object gen {
...
\ No newline at end of file
+    private def hostFunDefToFunction(gen: shine.OpenCL.HostCodeGenerator =
+                                        shine.OpenCL.HostCodeGenerator()
+                                    ): HostFunDef => CModule =
+      _.translateToModule(gen)
+
+    private def partialHostCompiler(hostFunName: String): PartialCompiler[
+      Phrase, HostedModule,
+      (HostFunDef, Seq[SizedKernelDef]),
+      (CModule, Seq[KernelModule])] =
```
Bastacyclop, line 203: maybe we can put a comment on `type PartialCompiler[S, T, S_, T_]` saying that it reduces compiling `S` to `T` to compiling `S_` to `T_`?

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
```diff
@@ -0,0 +1,200 @@
...
+{
+  override def name: String = "OpenCL Host"
+
+  override def cmd(env: Environment): Phrase[CommType] => Stmt = {
+    case KernelCallCmd(name, LocalSize(ls), GlobalSize(gs), output, args) =>
+      val arg_count = 1 + args.size
+      output |> acc(env, Nil, outputC =>
+        expSeq(args, env, argsC =>
+          C.AST.Block(
+            Seq(
```
michel-steuwer, line 31: Some comments explaining the individal code lines/blocks that get generated would be helpful.

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
```diff
@@ -0,0 +1,200 @@
...
+}
+
+case class HostCodeGenerator(override val decls: C.CodeGenerator.Declarations,
+                             override val ranges: C.CodeGenerator.Ranges)
+  extends C.CodeGenerator(decls, ranges)
+{
+  override def name: String = "OpenCL Host"
+
+  override def cmd(env: Environment): Phrase[CommType] => Stmt = {
+    case KernelCallCmd(name, LocalSize(ls), GlobalSize(gs), output, args) =>
```
michel-steuwer, line 26: maybe these cases are soo big that it makes sense to push them into individual functions?

src/main/scala/shine/OpenCL/SeparateHostAndKernelCode.scala
```diff
@@ -0,0 +1,127 @@
...
+import shine.DPIA.Types._
+import shine.OpenCL.primitives.functional.{KernelCall, Run}
+import shine.OpenCL.primitives.imperative.OpenCLKernelDefinition
+import scala.collection.mutable
+
+import scala.language.existentials
+import scala.annotation.tailrec
+
+object SeparateHostAndKernelCode {
+  type SizedKernelDef = (LocalSize, GlobalSize, OpenCLKernelDefinition)
```
michel-steuwer, line 15: As this is part of the interface (and therefore not a local type) I think it would make sense to put this elsewhere and make it a case class (or to add an Option for Local/Global size to `OpenCLKernelDefinition`.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
```diff
@@ -0,0 +1,271 @@
...
+import shine.DPIA.Types._
+import shine.DPIA.primitives.{imperative => dpia}
+import shine.OpenCL._
+import shine.OpenCL.primitives.{imperative => ocl}
+
+import scala.annotation.tailrec
+import scala.collection.{immutable, mutable}
+
+object HostManagedBuffers {
+  def populate(params: immutable.Seq[Identifier[ExpType]], outParam: Identifier[AccType])
```
michel-steuwer, line 15: Can you give a brief comment explaining the behaviour of this function. What is populated with what?

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
```diff
@@ -0,0 +1,271 @@
...
+  : Phrase[CommType] => Phrase[CommType] = { p =>
+    def outsideParams() = mutable.Set[Identifier[_ <: PhraseType]]() ++ (params :+ outParam)
+    val worstCasePrevious =
```
michel-steuwer, line 18: Can you explain the terminology here? Why previous?

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
```diff
@@ -0,0 +1,271 @@
...
+      Metadata(outsideParams(), outsideParams(), outsideParams(), outsideParams())
+    val managed = mutable.Map[Identifier[_ <: PhraseType], AccessFlags]()
+    insertHostExecutions(worstCasePrevious, outsideParams().toSet, managed, p)._1 |>
+    insertManagedBuffers(managed)
+  }
+
+  final case class HostExecution(env: Map[Identifier[_ <: PhraseType], AccessFlags],
```
michel-steuwer, line 25: What is a `HostExecution`? Why is this part of the `HostManagedBuffers` object?

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
```diff
@@ -0,0 +1,271 @@
...
+  private def insertManagedBuffers(
+    managed: mutable.Map[Identifier[_ <: PhraseType], AccessFlags]
+  ): Phrase[CommType] => Phrase[CommType] = p => {
+    val managed2 = managed.iterator.flatMap { case (i, a) =>
+      optionallyManaged(i).map(om => i -> (a, om._1))
+    }.toMap
+    VisitAndRebuild(p, Visitor2(managed2))
+  }
+
+  private case class Visitor2(
```
michel-steuwer, line 151: Can we try to find more expressive names for this (and the Visitor above)?

src/test/scala/shine/host.scala
```diff
@@ -0,0 +1,132 @@
...
+    findDeviceBufferSyncWrite(1, hostCode)
+    findDeviceBufferSyncRead(1, hostCode)
+    findCount(1, """hostBufferSync\(.*, HOST_WRITE\)""".r, hostCode)
+    findCount(1, """hostBufferSync\(.*, HOST_READ\)""".r, hostCode)
+    checkOutput(m)
+  }
+
+  test("two kernel calls") {
+    val e = depFun((n: Nat) => fun((n`.`i32) ->: (n`.`i32))(in =>
+      oclRun(LocalSize(2), GlobalSize(n/2))(mapGlobal(add(li32(1)))(in)) |> store(x =>
```
michel-steuwer, line 102: Do I understand correctly that every `oclRun` has to be followed by a `toMem` /  `store` when there is further computation following (i.e., it is not the last computation in the entrie RISE program)?

michel-steuwer (commented):
Thanks, @bastian-koepcke for this extensive PR.

Overall this looks great and will greatly improve the capabilities of RISE!

I have added some comments for you to look at and comment on.
And finally, some additional general points:
 - Is there a good reason not to include the OpenCL headers in the runtime? As the headers are standardized I can't see any downside and see the upside of having one less dependency to configure.
  - Can we add a description of the environment variables that have to be set to make this compile in the `docs/setup.md` document?

src/main/scala/rise/openCL/primitives.scala
```diff
@@ -64,4 +64,12 @@ object primitives {
         (s ->: s) ->: // function to write a value
           ((n + sz)`.`s) ->: ((1 + n)`.`sz`.`s) })})
   }
+
+  @primitive object oclRunP extends Primitive with Builder {
```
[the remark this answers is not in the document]
Bastacyclop: `P` stands for primitive, this is mainly to avoid name clashing with the `oclRun` DSL function.

src/main/scala/shine/OpenCL/Module.scala
```diff
@@ -0,0 +1,38 @@
...
+      kernels ++ other.kernels)
+}
+
+object Module {
+  def compose(ms: Seq[Module]): Module = ms.reduce(_ compose _)
+
+  def translateToString(m: Module): String =
+    s"""
+       |${m.kernels.map { km =>
+      // assumes a single kernel per module
```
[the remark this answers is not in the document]
Bastacyclop: I am not sure if we actually want this long-term? it might actually be beneficial to bundle all the kernels into a single module so that common definitions can be shared instead of duplicated? This would require changing the runtime and codegen slightly, so this was the easiest short-term solution.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
Bastacyclop: A `HostExecution` is doing some processing on the host and basically can be delegated to the plain C code generator without being aware of the host/device setup. I do not have a very good name in mind but I see it as launching a C kernel on the host (in constrast to launching an OpenCL kernel on the device).

src/test/scala/shine/host.scala
Bastacyclop: Yes, `oclRun` is always `write` and does not allocate memory, `toMem` allocates memory and goes from `write` to `read`.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
Bastacyclop: and the primitive can be extracted from the `HostManagedBuffers` object if/when the name is satisfying, this happened while coding as I first planned to eliminate this primitive before codegen.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
Bastacyclop: `Metadata` represents effects, these are the effects that could have happened previously: before the host function was called. I can add a comment.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
Bastacyclop: `InsertHostExecutionsVisitor`/`InsertManagedBuffersVisitor` modulo terminology changes we might do.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
Bastacyclop: What about: "Given a host program which only operates on plain arrays, this pass populates host managed buffers where needed, as well as host execution sections where plain arrays can still be used."?

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
Bastacyclop: I can do that, although I am not convinced that it would be more readable.

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
Bastacyclop: Will do.

src/main/scala/shine/OpenCL/HostFunctionDefinition.scala
```diff
@@ -0,0 +1,112 @@
...
+import shine.DPIA.Types._
+import shine.DPIA._
+import shine.DPIA.primitives.functional
+import shine.OpenCL.compilation.HostManagedBuffers
+import shine.{C, DPIA}
+import util.compiler.DSL.run
+
+import scala.collection.immutable
+
+final case class HostFunctionDefinition(name: String,
```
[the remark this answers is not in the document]
Bastacyclop: Conceptually the main difference is the function signature, a C function will use plain C arrays in its signature while a host function will take managed buffers in its signature. This means that the `HostManagedBuffers` pass is called while rewriting to imperative and that function parameters are modified when generating code and bundling up the module.

src/main/scala/shine/DPIA/InferAccessAnnotation.scala
```diff
@@ -134,6 +134,19 @@ private class InferAccessAnnotation {
         inferDepApp(depA, ctx, addsKernelParam(e, isKernelParamFun))
       case p: r.Primitive => inferPrimitive(p)
     }
+    // the kernel output should be 'write'
```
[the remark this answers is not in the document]
Bastacyclop: will change to must

src/main/scala/shine/DPIA/Compilation/CodeGenerator.scala
```diff
@@ -48,13 +48,17 @@ trait CodeGenerator {
 object CodeGenerator {
   final case class
     Environment(identEnv: immutable.Map[Identifier[_ <: BasePhraseType], C.AST.DeclRef],
+                varEnv: immutable.Map[Identifier[VarType], C.AST.DeclRef],
```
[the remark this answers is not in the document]
Bastacyclop: I am not 100% sure that it cannot be eliminated, but right now it is required because depending on the context, a managed buffer might be used as a regular acceptor/expression, or as a full variable, and can be represented with an acceptor/expression identifier (input/output) or a variable identifier (temporary). In any case the buffer use should generate the same C identifier.

src/main/scala/shine/C/AST/package.scala
```diff
@@ -6,9 +6,22 @@ import shine.DPIA.Types._
...
+    override def toString: String = s"#include <$name>"
+  }
+  case class IncludeSource(path: String) extends IncludeDirective {
+    override def toString: String = s"#include ${'"'}$path${'"'}"
+  }
+
   def makeParam(gen: CodeGenerator)(i: Identifier[_]): C.AST.ParamDecl = {
     // Turn array types into pointer types
     val paramType = getDataType(i) match {
+      // TODO: this function should not know about Buffer?
```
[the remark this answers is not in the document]
Bastacyclop: I can move the extra cases to `HostFunctionDefinition`.

src/main/scala/rise/openCL/primitives.scala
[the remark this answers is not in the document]
michel-steuwer: can we rename this to `oclRunPrimitive` (or something else)?

src/main/scala/shine/DPIA/Compilation/CodeGenerator.scala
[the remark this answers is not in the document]
michel-steuwer: Could we combine this with the `identEnv` map to allow identifiers with a var type to appear in them as well (i.e., relaxing the type of the `identEnv` map)?

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
michel-steuwer: At least it helps to avoid deep indentation

src/main/scala/rise/openCL/primitives.scala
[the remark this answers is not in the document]
Bastacyclop: coming in my next commit

src/main/scala/shine/DPIA/fromRise.scala
```diff
@@ -800,6 +800,18 @@ object fromRise {
             ocl.ToMem(a, t, e)))
       }
 
+      case rocl.oclRunP() => fromType {
```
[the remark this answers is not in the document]
Bastacyclop: I think that it is our current style, like we have `oclReduceSeq` in RISE and `ocl.ReduceSeq` in DPIA.

src/main/scala/shine/OpenCL/package.scala
```diff
@@ -6,6 +6,11 @@ import shine.DPIA.Nat
 import scala.language.implicitConversions
 
 package object OpenCL {
+  type AccessFlags = Int
```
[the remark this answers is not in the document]
Bastacyclop: I do not know how to represent bitflags using scala enums, my understanding is that I would still need to provide concrete values and union etc operators myself, creating a lot of boilerplate. Maybe I could instead create a simple type wrapper if we want to have stronger typing, e.g. `case class AccessFlags(Int)`?

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
Bastacyclop: I started by using more temporary values whose name might be enough to clarify the intent, I can also add comments if it remains hard to read.

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
Bastacyclop: doing it

src/main/scala/shine/DPIA/Compilation/CodeGenerator.scala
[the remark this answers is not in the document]
Bastacyclop: We have code such as `C.CodeGenerator.applySubstitutions` that relies on the `_ <: BasePhraseType` bound, but I can try to relax it and fix all uses that rely on this bound.

Bastacyclop:
> Is there a good reason not to include the OpenCL headers in the runtime? As the headers are standardized I can't see any downside and see the upside of having one less dependency to configure.

I have always seen projects using OpenCL rely on the headers being installed, and all "getting started with OpenCL" instructions telling you to install the headers as well. It seems standard for me to expect both the OpenCL library and its headers to be part of the setup (or if we don't expect standard header location, we should not expect standard library location either). However I am not strongly against including the headers either.

> Can we add a description of the environment variables that have to be set to make this compile in the docs/setup.md document?

I will add some notes about setting `LIBRARY_PATH`, `LD_LIBRARY_PATH` and `CPATH` variables if the library or header location is not standard.

Bastacyclop:
PS: in fact, we could use the headers from `lib/executor`?

src/main/scala/shine/DPIA/Compilation/CodeGenerator.scala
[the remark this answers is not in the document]
Bastacyclop: I removed `varEnv` altogether, I noticed that it was never actually used for anything anymore (only updated).

src/main/scala/shine/OpenCL/SeparateHostAndKernelCode.scala
Bastacyclop: I added the size information to the kernel definition.

src/main/scala/shine/OpenCL/package.scala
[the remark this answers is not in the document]
michel-steuwer: Not a big deal. Let's keep it like it is. I think Scala 3 actually might have some features here: https://dotty.epfl.ch/docs/reference/other-new-features/opaques.html and https://dotty.epfl.ch/docs/reference/enums/enums.html

src/main/scala/shine/OpenCL/Module.scala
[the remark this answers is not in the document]
michel-steuwer: I guess my point was, that we will have to think about this and touch this piece of code then anyway.
Currently the types suggest that we could support multiple kernels, but we clearly don't (like here).

I am fine keeping it as a comment here.

src/main/scala/shine/OpenCL/HostFunctionDefinition.scala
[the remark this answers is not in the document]
michel-steuwer: I have refactored the `ModuleGenerator` API and removed this file. See more for a comment below.

src/main/scala/shine/OpenCL/HostCodeGenerator.scala
michel-steuwer: :+1: I think that's fine.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
michel-steuwer: Thanks for adding the comment.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
michel-steuwer: Can we maybe try to replace the name `Metadata` as it is very generic and not helpful in the understanding of the code?
Maybe: `ReadsAndWrites`?

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
michel-steuwer: Thanks for pulling this out and adding an explaining comment.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
michel-steuwer: :+1:

src/test/scala/shine/host.scala
michel-steuwer: Thanks for the confirmation.

michel-steuwer:
I have restructured the `ModuleGenerator` APIs by:
 - introducing a generic trait `ModuleGenerator` that provides a default implementation for generating a Module (that could be a `C.Module`, an `OpenCL.Module` (i.e. a module mixing C code and Kernels), or a `OpenCL.KernelModule`) from a Function (Or Kernel) definition in multiple stages.
 - The individual `ModuleGenerator`s now overload the individual stages: `rewriteToImperative`, `imperativePasses`, `generateCode` and `makeModule`
 - This separates the module generation code from the function / kernel definition
 - This design simplifies reuse of individual stages
 - I have also changes the package structure slightly (e.g. introduced a `Compilation.Passes` package)

src/main/scala/shine/DPIA/Compilation/ModuleGenerator.scala
```diff
@@ -0,0 +1,52 @@
...
+  type GeneratedCode
+  def generateCode(gen: CodeGenerator,
+                   funDef: FunDef,
+                   outParam: Identifier[AccType]
+                  ): PhraseAfterPasses => GeneratedCode
+
+  def makeModule(gen: CodeGenerator,
+                 funDef: FunDef,
+                 outParam: Identifier[AccType]
+                ): GeneratedCode => Module
```
Bastacyclop, line 51: is it useful to separate `imperativePasses`, `generateCode` and `makeModule` here? I wonder if the `ModuleGenerator` API should not merge them into something like `imperativeToModule`, avoiding these arbitrary stages with abstract result types. Specific module generators can split this process into stages as they desire?

Bastacyclop (commented):
Looks good to me overall, although I feel like the module generator stages are over-specified/over-detailed in the trait.

src/main/scala/shine/C/Compilation/CodeGenerator.scala
```diff
@@ -1,18 +1,19 @@
-package shine.C
+package shine.C.Compilation
 
 import arithexpr.arithmetic.BoolExpr.ArithPredicate
 import arithexpr.arithmetic.{NamedVar, _}
+import shine.C.AST
 import shine.C.AST.Block
 import shine.C.AST.Type.getBaseType
-import shine.DPIA.Compilation.SimplifyNats
+import shine.DPIA.Compilation.Passes.SimplifyNats
```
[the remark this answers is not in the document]
michel-steuwer: I will fix this right here and we should have a separate PR at some point to rename all packages.

src/main/scala/shine/OpenCL/Compilation/HostCodeCodeGenerator.scala
```diff
@@ -1,24 +1,25 @@
...
-import shine.OpenCL.compilation.HostManagedBuffers
 import shine.OpenCL.primitives.imperative.{HostExecution, KernelCallCmd, NewManagedBuffer}
+import shine.OpenCL._
 
 import scala.collection.{immutable, mutable}
 
-object HostCodeGenerator {
-  def apply(): HostCodeGenerator =
-    new HostCodeGenerator(mutable.ListBuffer[Decl](), immutable.Map[String, arithmetic.Range]())
+object HostCodeCodeGenerator {
```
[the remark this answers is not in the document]
michel-steuwer: This is very much deliberate.
We generate `HostCode`, therefore, it is a `(HostCode)(CodeGenerator)`.
But, I would be ok reverting this (I think at some point I wanted to avoid a clash with the `HostModuleGenerator` which had a different name at some point ...
What do you think?

src/main/scala/shine/OpenCL/Compilation/Passes/InjectWorkItemSizes.scala
```diff
@@ -7,22 +7,17 @@ import shine.DPIA.Types.{CommType, DataType}
...
-
-        val gSizes: NDRange = if (globalSize.isEmpty) (get_global_size(0), get_global_size(1), get_global_size(3))
-        else (globalSize.get.size(0), globalSize.get.size(1), globalSize.get.size(2))
-
-        val numGroups: NDRange = if (localSize.isEmpty || globalSize.isEmpty)
-          (get_num_groups(0), get_num_groups(1), get_num_groups(2))
-        else (globalSize.get.size(0) /^ localSize.get.size(0),
-          globalSize.get.size(1) /^ localSize.get.size(1),
-          globalSize.get.size(2) /^ localSize.get.size(2))
+  def inject(wgConfig: Option[(LocalSize, GlobalSize)]
```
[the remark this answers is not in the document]
michel-steuwer: That wasn't supported before either. If we introduce this option, I am fine with having to change this interface.

src/main/scala/shine/OpenCL/Compilation/Passes/InsertMemoryBarriers.scala
```diff
@@ -17,27 +17,25 @@ object InsertMemoryBarriers {
...
-  private def visitLoopBody(
-    p: Phrase[CommType],
-    allocs: Map[Identifier[_ <: PhraseType], AddressSpace],
-    metadata: Metadata,
-    outer_wg_writes: mutable.Map[Identifier[_ <: PhraseType], AddressSpace] = mutable.Map()
-  ): Phrase[CommType] = {
+  private def visitLoopBody(p: Phrase[CommType],
+                            allocs: Map[Identifier[_ <: PhraseType], AddressSpace],
+                            metadata: Metadata,
+                            outer_wg_writes: mutable.Map[Identifier[_ <: PhraseType], AddressSpace] = mutable.Map()
```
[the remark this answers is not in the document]
michel-steuwer: Yeah, I know that you are not a fan of this style, but it's pretty much everywhere.
I will check if the lines are longer than 100 chars, otherwise, I will keep it.

src/main/scala/shine/OpenCL/Compilation/SeparateHostAndKernelCode.scala
```diff
@@ -49,7 +47,7 @@ object SeparateHostAndKernelCode {
         case _ => Continue(p, this)
       }
     })
-    (HostFunctionDefinition(hostFunName, hostDefinition), kernelDefinitions.toSeq)
+    (FunDef(hostFunName, hostDefinition), kernelDefinitions.toSeq)
```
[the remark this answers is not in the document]
michel-steuwer: One of the changes in the commit is that there is no longer a special `HostFunDef`, we are reusing the same `FunDef` as in normal C, we just generate a different Module from it.

src/main/scala/shine/OpenCL/Module.scala
```diff
@@ -1,26 +1,33 @@
 package shine.OpenCL
 
-case class Module(host: shine.C.Module, kernels: Seq[KernelModule]) {
+import shine.C
+
+// An OpenCL Module consists of the _host code_ a set of OpenCL _kernels_
```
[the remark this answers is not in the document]
michel-steuwer: I'll add the and

src/main/scala/util/ExecuteOpenCL.scala
```diff
@@ -6,8 +6,11 @@ import scala.sys.process._
 object ExecuteOpenCL {
   case class Exception(msg: String) extends Throwable
 
+  val clHeaderPath = "lib/executor/lib/Executor/include/"
   val runtimePath = "runtime/"
+  val executorHeadersPath = "lib/executor/lib/Executor/include/"
   val libs = "-lm -lOpenCL"
+  val includes = s"-I$runtimePath -I$executorHeadersPath"
```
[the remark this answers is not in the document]
michel-steuwer: I will have a look and fix this.

src/main/scala/util/TupleOps.scala
```diff
@@ -0,0 +1,120 @@
...
+    def ++[K, L, M, N, O, P](y: (K, L, M, N, O, P)): (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P) = (x._1, x._2, x._3, x._4, x._5, x._6, x._7, x._8, x._9, x._10, y._1, y._2, y._3, y._4, y._5, y._6)
+    def ++[K, L, M, N, O, P, Q](y: (K, L, M, N, O, P, Q)): (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q) = (x._1, x._2, x._3, x._4, x._5, x._6, x._7, x._8, x._9, x._10, y._1, y._2, y._3, y._4, y._5, y._6, y._7)
+    def ++[K, L, M, N, O, P, Q, R](y: (K, L, M, N, O, P, Q, R)): (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R) = (x._1, x._2, x._3, x._4, x._5, x._6, x._7, x._8, x._9, x._10, y._1, y._2, y._3, y._4, y._5, y._6, y._7, y._8)
+    def ++[K, L, M, N, O, P, Q, R, S](y: (K, L, M, N, O, P, Q, R, S)): (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S) = (x._1, x._2, x._3, x._4, x._5, x._6, x._7, x._8, x._9, x._10, y._1, y._2, y._3, y._4, y._5, y._6, y._7, y._8, y._9)
+    def ++[K, L, M, N, O, P, Q, R, S, T](y: (K, L, M, N, O, P, Q, R, S, T)): (A, B, C, D, E, F, G, H, I, J, K, L, M, N, O, P, Q, R, S, T) = (x._1, x._2, x._3, x._4, x._5, x._6, x._7, x._8, x._9, x._10, y._1, y._2, y._3, y._4, y._5, y._6, y._7, y._8, y._9, y._10)
+  }
+
+  // generator
```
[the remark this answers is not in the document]
michel-steuwer: No. I just kept the function that I wrote to generate the code above. I will comment it out.

src/test/scala/shine/host.scala
```diff
@@ -61,7 +61,7 @@ int main(int argc, char** argv) {
       mapSeq(add(li32(3)))(in)
     ))
     val m = gen.opencl.hosted.fromExpr(e)
-    val hostCode = gen.c.function.asString(m.host)
+    val hostCode = gen.c.function.asString(m.hostCode)
```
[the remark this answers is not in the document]
michel-steuwer: We can not use `host` to refer to computation (in whatever form). The host is the machine executing part of an OpenCL program. I think the term `hostCode` is perfectly acceptable to refer to the part of the OpenCL module that is executed on the host.

src/main/scala/shine/OpenCL/Compilation/SeparateHostAndKernelCode.scala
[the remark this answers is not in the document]
Bastacyclop: can we use it as `C.FunDef` then?

src/main/scala/shine/DPIA/Compilation/ModuleGenerator.scala
michel-steuwer: One advantage is that we can independently refer (and invoke) the passes that are performed, e.g., by the `C.ModuleGenerator`.
This way it is easy to ensure that the e.g. `OpenCL.KernelGenerator` is performing the same passes (plus more) as the C generator.

If we mix them all up in one big step this aspect becomes harder.

src/main/scala/shine/OpenCL/compilation/HostManagedBuffers.scala
michel-steuwer: I have renamed `Metadata` to `ReadsAndWrites`

src/main/scala/shine/C/Compilation/CodeGenerator.scala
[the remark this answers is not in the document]
michel-steuwer: Well, actually. Let's fix this in a separate PR.

src/main/scala/shine/OpenCL/Compilation/HostCodeCodeGenerator.scala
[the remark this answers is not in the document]
Bastacyclop: I prefer `HostCodeGenerator`, it is not a code generator of host code, it is simply a generator of host code.

src/main/scala/shine/DPIA/Compilation/ModuleGenerator.scala
Bastacyclop: I agree that `OpenCL.KernelGenerator` wants to refer to some `C.ModuleGenerator` passes, but this does not require any trait. I believe that the trait should only contain methods that are useful/meaninful in the generic context. It does not look like anybody would use `mg: ModuleGenerator; mg.imperativePasses()`. They would not even know what it returns concretely. We can keep these methods inside the concrete generators and remove them from the generic trait?

src/main/scala/shine/OpenCL/Compilation/HostManagedBuffers.scala
```diff
@@ -0,0 +1,265 @@
...
+
+
+  // effects to outer scope allocations
+  private case class ReadsAndWrites(
+    host_reads: mutable.Set[Identifier[_ <: PhraseType]],
+    host_writes: mutable.Set[Identifier[_ <: PhraseType]],
+    device_reads: mutable.Set[Identifier[_ <: PhraseType]],
+    device_writes: mutable.Set[Identifier[_ <: PhraseType]])
+
+  private object Metadata {
```
Bastacyclop, line 36: this should also be `ReadsAndWrites`

michel-steuwer: yep. good catch. I'll fix that.

src/main/scala/util/ExecuteOpenCL.scala
```diff
@@ -42,8 +61,8 @@ object ExecuteOpenCL {
     try {
       val src = writeToTempFile("code-", ".c", code).getAbsolutePath
       val bin = createTempFile("bin-", "").getAbsolutePath
-      val sources = s"$src $runtimePath/buffer_${buffer_impl}.c $runtimePath/ocl.c"
-      (s"clang -O2 $sources $includes -o $bin $libs -Wno-parentheses-equality" !!)
+      val sources = s"$src $runtimePath/buffer_$buffer_impl.c $runtimePath/ocl.c"
```
Bastacyclop: its a bit hard to read that this is interpolating `buffer_impl`?

michel-steuwer: For me, the syntax highlighting helps to clarify this

src/main/scala/shine/DPIA/Compilation/ModuleGenerator.scala
michel-steuwer: I see your point and https://github.com/rise-lang/shine/pull/92/commits/77cf7c004a082fb1ec3de42ff21e82a0f254def0 provides a simplified version of the ModuleGenerator trait.

PR #108, opened 2021-02-12
Implement `hashCode` for `Expr` and `Type`

### LINKED ISSUE #104: Make `Expr.hashCode` follow `Expr.equals`
`Expr.hashCode` tests for syntactic equality, while `Expr.equals` tests for alpha-equivalence. Discriminating by `Expr.hashCode` is only ever used to then discriminate by `Expr.equals`, the former should be more liberal.
A more comprehensive solution is discussed in #77.

umazalakain:
Closes #104

### DIFF
--- a/src/main/scala/rise/core/Expr.scala
+++ b/src/main/scala/rise/core/Expr.scala
@@ -1,5 +1,6 @@
 package rise.core
 
+import semantics._
 import rise.core.types._
 import rise.core.DSL.Type.TypeEqual
 import rise.core.ShowRise._
@@ -8,6 +9,19 @@ sealed abstract class Expr {
   val t: Type
   def setType(t: Type): Expr
   override def toString: String = showRise(this)
+  override def hashCode(): Int = this match {
+    case _: Identifier => 17
+    case Lambda(_, e) => 3 * e.hashCode() + 1
+    case App(f, e) => 5 * f.hashCode() + -7 * e.hashCode() + 2
+    case DepLambda(_, e) => 4 * e.hashCode() + 3
+    case DepApp(f, _) => 6 * f.hashCode() + 4
+    case l@Literal(_: ScalarData | _: VectorData) => l.d.hashCode()
+    case Literal(_: NatData) => 91
+    case Literal(_: IndexData) => 93
+    case Literal(_: ArrayData) => 95
+    case Literal(_: PairData) => 97
+    case p: Primitive => p.getClass.hashCode()
+  }
 }
 
 final case class Identifier(name: String)(
--- a/src/main/scala/rise/core/Expr.scala
+++ b/src/main/scala/rise/core/Expr.scala
@@ -6,28 +6,30 @@ import rise.core.DSL.Type.TypeEqual
 import rise.core.ShowRise._
 
 object alphaEquiv {
-  val equiv : Expr => Expr => Boolean = a => b => {
-    a.t =~= b.t && ((a, b) match {
+  type TYPEEQ = Type => Type => Boolean
+
+  val equiv : TYPEEQ => Expr => Expr => Boolean = typeEq => a => b => {
+    typeEq(a.t)(b.t) && ((a, b) match {
       case (Identifier(na), Identifier(nb)) => na == nb
       case (Literal(da), Literal(db)) => da == db
       case (a : Primitive, b : Primitive) => a.name == b.name
       // Application compares structurally
-      case (App(fa, ea), App(fb, eb)) => equiv(fa)(fb) && equiv(ea)(eb)
-      case (DepApp(fa, xa), DepApp(fb, xb)) => equiv(fa)(fb) && xa == xb
+      case (App(fa, ea), App(fb, eb)) => equiv(typeEq)(fa)(fb) && equiv(typeEq)(ea)(eb)
+      case (DepApp(fa, xa), DepApp(fb, xb)) => equiv(typeEq)(fa)(fb) && xa == xb
       // Abstraction compares after substitution
       case (Lambda(xa, ta), other@Lambda(xb, _)) =>
-        xa.t =~= xb.t && equiv(ta)(typedLifting.liftFunExpr(other).value(xa))
+        xa.t =~= xb.t && equiv(typeEq)(ta)(typedLifting.liftFunExpr(other).value(xa))
       case (DepLambda(xa, ea), other@DepLambda(xb, _)) => (xa, xb) match {
         case (n: NatIdentifier, _: NatIdentifier) =>
-          equiv(ea)(typedLifting.liftDepFunExpr[NatKind](other).value(n))
+          equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[NatKind](other).value(n))
         case (dt: DataTypeIdentifier, _: DataTypeIdentifier) =>
-          equiv(ea)(typedLifting.liftDepFunExpr[DataKind](other).value(dt))
+          equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[DataKind](other).value(dt))
         case (addr: AddressSpaceIdentifier, _: AddressSpaceIdentifier) =>
-          equiv(ea)(typedLifting.liftDepFunExpr[AddressSpaceKind](other).value(addr))
+          equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[AddressSpaceKind](other).value(addr))
         case (n2n: NatToNatIdentifier, _: NatToNatIdentifier) =>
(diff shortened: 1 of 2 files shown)


src/main/scala/rise/core/Expr.scala
```diff
@@ -8,6 +9,19 @@ sealed abstract class Expr {
   val t: Type
   def setType(t: Type): Expr
   override def toString: String = showRise(this)
+  override def hashCode(): Int = this match {
+    case _: Identifier => 17
```
Bastacyclop, line 13: I think that this might impact negatively all of our code which uses sets/maps of `Identifier`s? Should they map over the name instead to have a proper hash then?

Bastacyclop: can this change also impact code such as subsitutions?

Bastacyclop (commented):
I wonder if it would not be less error-prone to opt-in for alpha equivalent equality / hashing after all. Some of our code does care about identifier names (maps of ids, substitution, ..).

src/main/scala/rise/core/types/Type.scala
```diff
@@ -3,7 +3,15 @@ package rise.core.types
...
 import rise.core._
 
-sealed trait Type
+sealed trait Type {
+  override def hashCode(): Int = this match {
+    case TypePlaceholder => 5
+    case TypeIdentifier(n) => 7 * n.hashCode()
+    case FunType(inT, outT) => 11 * inT.hashCode() + 13 * outT.hashCode() + 1
+    case DepFunType(_, t) => 17 * t.hashCode()
+    case dataType: DataType => 19 * dataType.hashCode()
```
Bastacyclop, line 12: this does not seem to respect alpha-equivalence for `TypeIdentifier` and other identifiers found in `DataType`.

src/main/scala/rise/core/types/Type.scala
```diff
@@ -3,8 +3,29 @@ package rise.core.types
 import arithexpr.arithmetic.RangeAdd
 import rise.core._
 
-sealed trait Type {
-  override def hashCode(): Int = this match {
+object alphaEquiv {
+  val equiv : Type => Type => Boolean = a => b => {
```
umazalakain, line 7: @Bastacyclop  @fedepiz I'm trying to abstract away type equality with alpha equivalence here, while still keeping the old behavior for the time being by having `Type.equals` actually call this. I've done something similar with alpha equivalence for terms, and that works, but this one here doesn't, and I cannot pin point why. Am I missing anything obvious?

umazalakain: Got it! It was a missing case for pair types. Thinking of going for a more defensive programming style where the compiler can help with exhaustiveness checking.

src/main/scala/rise/core/Expr.scala
```diff
@@ -1,46 +1,82 @@
 package rise.core
 
+import semantics._
 import rise.core.types._
 import rise.core.DSL.Type.TypeEqual
 import rise.core.ShowRise._
 
+object alphaEquiv {
+  type TYPEEQ = Type => Type => Boolean
```
Bastacyclop, line 9: maybe `TypeEq`?

src/main/scala/rise/core/Expr.scala
```diff
@@ -1,46 +1,82 @@
...
+        case (n2d: NatToDataIdentifier, _: NatToDataIdentifier) =>
+          equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[NatToDataKind](other).value(n2d))
+        case _ => false
+      }
+      case _ => false
+    })
+  }
+
+  val hash : Expr => Int = {
+    case _: Identifier => 17
```
Bastacyclop, line 40: maybe in the future we can use some sort of DeBruijn indices to get a more useful hash, e.g. it would differentiate `\x. \y. x y` and `\x. \y. y x`. not sure if that it worth the effort but maybe we can add a comment about this in the code.

src/main/scala/rise/core/Expr.scala
```diff
@@ -1,46 +1,82 @@
...
+    case Lambda(_, e) => 3 * e.hashCode() + 1
+    case App(f, e) => 5 * f.hashCode() + -7 * e.hashCode() + 2
+    case DepLambda(_, e) => 4 * e.hashCode() + 3
+    case DepApp(f, _) => 6 * f.hashCode() + 4
+    case l@Literal(_: ScalarData | _: VectorData) => l.d.hashCode()
+    case Literal(_: NatData) => 91
+    case Literal(_: IndexData) => 93
+    case Literal(_: ArrayData) => 95
+    case Literal(_: PairData) => 97
+    case p: Primitive => p.getClass.hashCode()
```
Bastacyclop, line 50: `p.name.hashCode()` might allow to differentiate between e.g. `mapLocal(1)` and `mapLocal(2)`?

src/main/scala/rise/core/types/Type.scala
```diff
@@ -3,7 +3,47 @@ package rise.core.types
 import arithexpr.arithmetic.RangeAdd
 import rise.core._
 
-sealed trait Type
+object alphaEquiv {
+  val equiv : Type => Type => Boolean = a => b => {
+    (a.getClass == b.getClass && b.getClass == TypePlaceholder.getClass) ||
+    (a.getClass == b.getClass && b.getClass == NatType.getClass) ||
```
Bastacyclop, line 9: why not add these two cases in the `match` below?

src/main/scala/rise/core/types/Type.scala
```diff
@@ -3,7 +3,47 @@ package rise.core.types
...
+      ((a, b) match {
+      case (TypeIdentifier(na), TypeIdentifier(nb)) => na == nb
+      case (DataTypeIdentifier(na, _), DataTypeIdentifier(nb, _)) => na == nb
+      case (FunType(sa, ta), FunType(sb, tb)) => equiv(sa)(sb) && equiv(ta)(tb)
+      case (DepFunType(xa, ta), other@DepFunType(_, _)) =>
+        equiv(ta)(lifting.liftDependentFunctionType(other)(xa))
+      case (sa : ScalarType, sb : ScalarType) => sa.getClass == sb.getClass
```
Bastacyclop, line 16: why not `sa == sb`?

Bastacyclop (changes requested):

src/main/scala/rise/core/Expr.scala
umazalakain: Yeah, sounds better. Btw I'm thinking about moving all the equality stuff to `core/Equality` and `core/Types/Equality`, how does this sound?

Bastacyclop: I think `core.equality` sounds good to me and I would include the type equality there as well (there is not a lot of code, right?)

src/main/scala/rise/core/Expr.scala
umazalakain: As in, implement an on the fly conversion to de Bruijn? Maybe we could switch altogether (first rise and then DPIA), but yeah, it's worth documenting properly what this hash differentiates and what it doesn't.

src/main/scala/rise/core/Expr.scala
umazalakain: Didn't know about those, thanks!

src/main/scala/rise/core/types/Type.scala
umazalakain: This is a leftover from when `Type.equals` called `alphaEquiv.equiv`, as matching against objects calls `.equals` behind the scenes, so we were getting into infinite loops. I still need to rewrite alpha equivalence on terms so that it doesn't use substitution, like I did with types (potentially much bigger speed up? fingers crossed), I will do it as part of that.

umazalakain:
@Bastacyclop Sorry, I should have made up my mind earlier, but I'm thinking of closing this in favour of that WIP that rewrites alpha equivalence, how does that sound?

src/main/scala/rise/core/Expr.scala
umazalakain: No, there isn't, sounds good!

src/main/scala/rise/core/Expr.scala
Bastacyclop: Yeah, on-the-fly with a simple `Map[Name, Int]` where you start with index 0 and increase when you go into lambdas, although this might be too expensive for a hash, not sure. I am not convinced that we want to use DeBruijn indices in the entire codebase, I am more tempted by graph representations, but that is another question ;p.

Bastacyclop:
@umazalakain  sounds reasonable

PR #89, opened 2021-01-08
Add some installation instructions

umazalakain:
I have tested the instructions for NixOS. They depend on https://github.com/rise-lang/shine/pull/88 and https://github.com/rise-lang/executor/pull/8 being merged. I might make a nix derivation out of this, but for now I'm happy it works.

PR #109, opened 2021-02-15
Documentation website (including first Tutorial)

michel-steuwer:
Added Tutorial app showing a basic overview of RISE and ELEVATE being used together.

This PR adds some additional syntax and functions to make the ELEVATE site a bit easier to understand (in the context of the tutorial).

There are also two bug fixes included:
 - the default traversal order of ELEVATE programs now follows more closely the execution order making specifying traversals more intuitive
 - a bug in the OpenMP vectorization code


src/main/scala/rise/elevate/rules/traversal.scala
```diff
@@ -69,12 +69,24 @@ object traversal {
     implicit object RiseTraversable extends implementation.DefaultTraversal {
       override protected def oneHandlingState: Boolean => Strategy[Rise] => Strategy[Rise] =
         carryOverState => s => {
-          // (option 1) traverse to argument first
-          case a @ App(f, e) => s(e) match {
-            case Success(x: Rise) => Success(App(f, x)(a.t))
-            case Failure(state)   => if (carryOverState)
-              state(f).mapSuccess(App(_, e)(a.t)) else
-              s(f).mapSuccess(App(_, e)(a.t))
+          // To achieve a traversal that most closely corresponds to the execution order we ...
```
Bastacyclop, line 72: can you explain why we would want the traversal order to be closer to execution order?

src/main/scala/apps/Tutorial.scala
```diff
@@ -0,0 +1,200 @@
...
+      fun(n`.`k`.`f32)(A => fun(k`.`m`.`f32)(B =>
+        A |> map(fun(rowOfA =>
+          B |> transpose |> map(fun(colOfB =>
+            zip(rowOfA)(colOfB) |>
+              map(fun(x => fst(x) * snd(x))) |>
+              reduce(add)(l(0.0f)) )) )) )) )
+  // The matrix dimensions are represented as part of the type of the matrices:
+  //  - the type of matrix `A` is `n.k.f32`
+  //  - the type of matrix `B` is `k.m.f32`
+  // The identifies used in the type (here: `n`, `m`, and `k`) are introduced
```
Bastacyclop, line 45: identifiers

michel-steuwer:
This PR now also adds a documentation website, that is automatically deployed at: https://rise-lang.org/doc/

The process works as follows:
 - Every markdown file that is added to the `/docs/` folder corresponds to a page in the documentation
 - Scala code that is embedded in these markdown files is checked to compile and run using [mdoc](https://scalameta.org/mdoc/) (This is automatically run by the CI on every commit of a PR).
 - The documentation website is powered by [Docusaurus](https://v2.docusaurus.io) and configured in the `/docs-website/` folder. The documentation page is deployed automatically on every successfully merged PR.
 - The documentation website is hosted using GitHub pages on the `gh-pages` branch of the [`doc`](https://github.com/rise-lang/doc/tree/gh-pages) repository and publicly accessible at: https://rise-lang.org/doc/

michel-steuwer:
I have integrated the content from #89 in the documentation.
@umazalakain I, therefore, suggest to close #89 in favour of this PR.

docs-website/yarn.lock
```diff
@@ -0,0 +1,10318 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
Bastacyclop, line 2: this file is huge!

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,83 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  check-documentation:
+    if: github.event_name != 'push'
```
Bastacyclop, line 11: any reason not to use `== 'pull_request'`?

.github/workflows/documentation.yml
```diff
@@ -0,0 +1,83 @@
...
+          yarn install --frozen-lockfile
+          elif [ -e package-lock.json ]; then
+          npm ci
+          else
+          npm i
+          fi
+          npm run build
+
+  deploy-documentation:
+    if: github.event_name != 'pull_request'
```
Bastacyclop, line 39: any reason not to use `== 'push'`?

Bastacyclop:
@michel-steuwer Its great to have a documentation website! maybe we can combine this with API docs auto-generated from our source code in the future? (so docs embedded in our code as well as code embedded in our docs)

docs-website/yarn.lock
michel-steuwer: It's autogenerated and pins all dependencies. It is recommended to be checked in the version control so that the exact same version is used by the CI.

.github/workflows/documentation.yml
michel-steuwer: Yes. We still want to allow the `workflow_dispatch` (that is a manual trigger via the website) event to trigger this workflow.

.github/workflows/documentation.yml
michel-steuwer: See above.

src/main/scala/rise/elevate/rules/lowering.scala
```diff
@@ -29,34 +31,44 @@ object lowering {
     case _ => false
   }
 
+  def `map |-> mapSeq`: Strategy[Rise] = mapSeq
```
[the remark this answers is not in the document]
michel-steuwer: I will revert to `->` for now. I think we will have to make a separate push for consistency of the ELEVATE site of things at some point.

src/main/scala/rise/elevate/rules/traversal.scala
michel-steuwer: Otherwise, traversals such as `outermost` and `innermost` have not the excepted effect.

Without the proposed changes `outermost(isMap)` will trigger the most nested `map` in the following expression instead of the one in the first line:
```
A |> map(fun(arow =>
  B |> transpose |> map(fun(bcol =>
    zip(arow)(bcol) |> map(multP) |> reduce(add)(0)
  )
)
```

michel-steuwer:
@Bastacyclop thanks for the review. I have addressed all your comments. Could you have a final look and give your ok, or raise more points.

src/main/scala/rise/elevate/rules/traversal.scala
Bastacyclop: so if I traverse `f(g(v)) ~ v |> g |> f`, I will first traverse `f`, then `v`, then `g`? it is hard for me to see how this order fits the intuition of `outermost` or `innermost`, and more importantly why this should affect other traversals such as a simple `topDown` or `bottomUp`. What was the issue with the previous traversal order where we always visited the function first?

Bastacyclop (approved):

PR #110, opened 2021-02-16
Equality Saturation experiments

Bastacyclop:


Bastacyclop:
parts of this will be merged with https://github.com/rise-lang/shine/pull/114, the rest of the experiment is archived in https://github.com/rise-lang/shine/tree/eqsatCustom with no intent to merge into master.

Issue #35, opened 2020-01-29
Remove global counters


Bastacyclop:
I think that we should remove all global counters except the ones used in the rise DSL. For example during codegen we should use a program-wide counter instead of a compiler-wide counter. Currently generating code mulitple times keeps using bigger and bigger numbers, this is annoying for inspection and reproducibility and serves no purpose.

@rise-lang/core

PR #112, opened 2021-02-19
Add list of RISE primitives and their types to docs

umazalakain:
Exposes a list of primitives in `rise.core.primitives`, and uses mdoc to iterate over the list and print the `typeScheme` associated with each primitive. The type pretty printing is rather atrocious, not sure this is the best way to go about it.

### DIFF
--- a/docs/reference/rise-primitives.md
+++ b/docs/reference/rise-primitives.md
@@ -1,0 +1,30 @@
+---
+id: rise-primitives
+title: RISE Primitives
+sidebar_label: RISE Primitives
+slug: /reference/rise-primitives
+---
+
+## Type syntax cheatsheet
+
+**Note**: curly braces quantify over implicit parameters.
+
+Type | Syntax
+-----|-------
+Function from `A` to `B` | `(A -> B)`
+Dependent function from `a : A` to `B a` | `(x : A -> B)`
+Dependent function application | `f(n)`
+Vector of `A` of size `n` | `<n>A`
+Array of `A` of size `n` | `n.A`
+Dependent array of `F : Nat -> A` of size `n` | `n..F`
+Index of size `n : Nat` | `idx[n]`
+Pair of `A` and `B` | `(A, B)`
+Dependent pair of `x : A` and `B a` | `(x : A ** B)`
+
+## Primitives
+
+```scala mdoc:passthrough
+    val primTypeScheme = os.pwd / os.RelPath("src/main/scala/rise/core/primitives/primitives.rise")
+    val output = Seq("```", os.read(primTypeScheme), "```").mkString("\n")
+    println(output)
+```
--- a/docs-website/.gitignore
+++ b/docs-website/.gitignore
@@ -7,7 +7,6 @@
 # Generated files
 .docusaurus
 .cache-loader
-docs/*.md
 docs/**/*.md
 
 # Misc
--- a/docs-website/sidebars.js
+++ b/docs-website/sidebars.js
@@ -35,4 +35,9 @@ module.exports = {
       ],
     }
   ],
+  docs: {
+    'Getting started': ['greeting', 'setup'],
+    'Tutorials': ['overview'],
+    'Language Reference': ['reference/rise-primitives']
+  },
 };


docs/reference/rise-primitives.md
```diff
@@ -0,0 +1,37 @@
...
+import rise.core.{primitives, makeClosed}
+println("Primitive | Type")
+println("----------|-----")
+primitives.members.foreach { p =>
+  var ts = p.typeScheme
+  // Try and close the implicit identifiers in the types
+  try { ts = makeClosed(ts)._1 }
+    catch { case _ : Throwable => () }
+  // Prevent types from being interpreted as column separators
+  val tsrepr = ts.toString.replace("|", "")
```
michel-steuwer, line 34: This is a bit delicate. With this line here you change the meaning of some types.
Take `depZip` as an example. The type should be:
```
depZip: (n: nat -> (n.._n2d1 -> (n.._n2d2 -> n..(m: nat |-> (_n2d1(m), _n2d2(m))))))
                                                 ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~
                                                 This indicates that this is a type-level
                                                 anonymous function, not a dependent function type
```
Replacing the `|->` to a `->` changes the kind of the underlined expression.
Maybe we can find another notation to make clear that this arrow is different?

docs/reference/rise-primitives.md
```diff
@@ -0,0 +1,37 @@
...
+
+```scala mdoc:passthrough
+import rise.core.{primitives, makeClosed}
+println("Primitive | Type")
+println("----------|-----")
+primitives.members.foreach { p =>
+  var ts = p.typeScheme
+  // Try and close the implicit identifiers in the types
+  try { ts = makeClosed(ts)._1 }
+    catch { case _ : Throwable => () }
```
michel-steuwer, line 32: I am not 100% sure if I get why you are doing this?

What happens if the closing of the term silently fails?

docs/reference/rise-primitives.md
```diff
@@ -0,0 +1,37 @@
...
+Function from `A` to `B` | `(A -> B)`
+Dependent function from `a : A` to `B a` | `(x : A -> B)`
+Dependent function application | `f(n)`
+Vector of `A` of size `n` | `<n>A`
+Array of `A` of size `n` | `n.A`
+Dependent array of `F : Nat -> A` of size `n` | `n..F`
+Index of size `n : Nat` | `idx[n]`
+Pair of `A` and `B` | `(A, B)`
+Dependent pair of `x : A` and `B a` | `(x : A ** B)`
+
```
michel-steuwer, line 21: Can you add a note here that there is a convention that implicit parameters are written with a leading underscore?

michel-steuwer (commented):
Thanks, @umazalakain. That looks great.
Some questions for you to look at.

It would also be great if we could find a way to rename the identifiers before we print them.

docs/reference/rise-primitives.md
Bastacyclop: Is it possible to escape this symbol, using `&#124;` for example?

umazalakain: I tried a bunch of different things:
- Using `|` inside the code tags is still interpreted as "next table cell"
- Escaping with `\|` works outside of the code tags but not inside, as the backslash is automatically escaped, i.e. we get `\\|` that is rendered as `\|`.
- Escaping with `&#124;` works outside of the code tags but not inside, the ampersand is automatically escaped, i.e. we get `\&#124;` which is rendered as `&#124;`.
- Replacing `|->` with the unicode symbol `⟼` works but it looks kind of small.
- Replacing `|` with `⊢` looks good and gives `⊢->`.

docs-website/docs/reference/rise-primitives.md
```diff
@@ -0,0 +1,88 @@
+---
```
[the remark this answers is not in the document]
umazalakain: Missed that, thanks

docs/reference/rise-primitives.md
umazalakain: good idea

umazalakain: Mhm I just realised that because we use `makeClosed` to show the implicit variable bindings these get transformed into explicit variables and thus are not prefixed with an underscore anymore. Should we make the types closed and all bindings explicit or show we leave them open and not bind implicit variables but refer to them with underscores?

docs/reference/rise-primitives.md
umazalakain: In previous versions `makeClosed` broke for certain primitives, in which case it reverted to the unclosed version. However this seems not to be the case on current master anymore. There is still the question of whether we want closed expressions or not though.

umazalakain:
The need to rename identifiers is related to  #35. We could certainly have an ad-hoc solution here that uses its own fresh variable counter and does the renaming, but we might as well wait to solve it for every case.

docs/reference/rise-primitives.md
Bastacyclop: I think leaving the type open sounds good? It also reflects better what argument should really be passed when writing programs.

umazalakain: True, I'll leave them open :)

umazalakain: @Bastacyclop a small drawback to this is that we would not be printing out the types of these type identifiers anymore (unless we print out types together with the variable reference?)

Bastacyclop: I guess that you mean the kinds of the identifiers? The naming convention might be enough for this. Otherwise you could use `makeClosed` to spot the free variables and then print them differently. Maybe `forall n: Nat ...`.

Bastacyclop:
I'm not sure that fixing #35 would solve the name issue. These identifiers are created by the DSL which I think will be the only place where we might want to keep a global counter, otherwise every expression construction will need some local context parameter to get hold on a counter.

umazalakain:
Could that not be done by wrapping up the DSL calls? This could probably be automated in tests.

michel-steuwer:
@umazalakain should we close this, as we have now merged #127?
Is there something from here to bring into a new PR to bring the primitives into the documentation?

umazalakain:
@michel-steuwer pushed a basic printout of `primitives.rise` to the docs

PR #113, opened 2021-02-19
Refactor equality -- alpha renaming without substitution

umazalakain:
Leaves Scala's syntactic equality (for both terms and types) untouched -- this is the equality we want if we want to e.g. keep identifiers as keys in a map. Factors alpha equivalence out of the current `.equals` methods and into their own (`=~=`). Parametrises the equality on terms on an equality on types. Rewrites alpha equivalence on types and on expressions so that it can be checked without substitution, by keeping an environment of equivalent names.


src/main/scala/rise/elevate/rules/algorithmic.scala
```diff
@@ -90,7 +90,7 @@ object algorithmic {
   @rule def reduceMapFission()(implicit ev: Traversable[Rise]): Strategy[Rise] = {
     case e @ App(App(ReduceX(), Lambda(acc, Lambda(y,
         App(App(op, acc2), f@App(_, y2))))), init)
-      if acc == acc2 && contains[Rise](y).apply(y2) =>
+      if acc =~= acc2 && contains[Rise](y).apply(y2) =>
```
Bastacyclop, line 93: `=~~=`?

src/main/scala/rise/elevate/rules/algorithmic.scala
```diff
@@ -185,7 +185,7 @@ object algorithmic {
   }
 
   @rule def mapIdentity: Strategy[Rise] = {
-    case expr @ App(map(), Lambda(x1, x2)) if x1 == x2 => Success(fun(x => x) !: expr.t)
+    case expr @ App(map(), Lambda(x1, x2)) if x1 =~= x2 => Success(fun(x => x) !: expr.t)
```
Bastacyclop, line 188: `=~~=`?

src/main/scala/rise/elevate/rules/algorithmic.scala
```diff
@@ -250,7 +250,7 @@ object algorithmic {
       case App(makeArray(_), App(App(map(), _), e)) => Some(e)
       case App(f, App(App(map(), _), e2)) =>
         matchExpectedMakeArray(f).flatMap(e =>
-          if (e == e2) { Some(e) } else { None })
+          if (e =~= e2) { Some(e) } else { None })
```
Bastacyclop, line 253: `=~~=`?

src/main/scala/rise/elevate/rules/algorithmic.scala
```diff
@@ -275,7 +275,7 @@ object algorithmic {
     case expr @ App(generate(), Lambda(i, App(App(App(select(), t),
       App(App(map(), f), e1)),
       App(App(map(), g), e2))))
-    if e1 == e2 && !contains[Rise](i).apply(e1) =>
+    if e1 =~= e2 && !contains[Rise](i).apply(e1) =>
```
Bastacyclop, line 278: more of these, we should only rewrite fully typed expressions.

src/test/scala/apps/separableConvolution2DRewrite.scala
```diff
@@ -39,7 +39,7 @@ class separableConvolution2DRewrite extends test_util.Tests {
     val na = BENF(a).get
     val nb = BENF(b).get
     val uab: Rise = toBeTyped(na) !: nb.t
-    makeClosed(uab)._1 == makeClosed(nb)._1
+    makeClosed(uab)._1 =~= makeClosed(nb)._1
```
Bastacyclop, line 42: `=~~=`?

src/main/scala/rise/core/types/Type.scala
```diff
@@ -2,8 +2,88 @@ package rise.core.types
...
+  def makeExplicit[K <: Kind.Identifier] : K => K = {
+    case t : DataTypeIdentifier => t.asExplicit.asInstanceOf[K]
+    case t => t
+  }
+
+  def equiv : Type => Type => Boolean = equiv(Nil)
+
+  /** Alpha equivalence on types
+    * Datatype identifier explicitness is ignored.
+    * Short circuits in the event of syntactic equality.
```
Bastacyclop, line 25: is this short-circuit not more costly in most cases (we might have to do two traversals instead of one)?

src/main/scala/rise/core/Expr.scala
```diff
@@ -1,46 +1,81 @@
...
+            equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[DataKind](other).value(dt)) }
+          case addr: AddressSpaceIdentifier => and { case _: AddressSpaceIdentifier =>
+            equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[AddressSpaceKind](other).value(addr)) }
+          case n2n: NatToNatIdentifier => and { case _: NatToNatIdentifier =>
+            equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[NatToNatKind](other).value(n2n)) }
+          case n2d: NatToDataIdentifier => and { case _: NatToDataIdentifier =>
+            equiv(typeEq)(ea)(typedLifting.liftDepFunExpr[NatToDataKind](other).value(n2d)) }
+      }}
+    })
+  }
```
Bastacyclop, line 39: it would be good to avoid substitution for expressions as well, you will then have to communicate the environment for type equality.

src/main/scala/rise/elevate/rules/algorithmic.scala
umazalakain: The original behavior of == was to accept type placeholders. I will try and see which ones I can get rid of though.

src/main/scala/rise/core/Expr.scala
umazalakain: Agreed! It's the next item on my todo ^^

src/main/scala/rise/core/types/Type.scala
umazalakain: I need to do some benchmarking, but I suspect Scala's equals is heavily optimized, like, maybe it's able to see that it hasn't been overwritten and use a direct memory comparison or something.

src/main/scala/rise/core/types/Type.scala
```diff
@@ -2,8 +2,88 @@ package rise.core.types
...
+        xa.getClass == xb.getClass && equiv((makeExplicit(xa), makeExplicit(xb)) :: env)(ta)(tb) }
+    })
+  }
+
+  val hash : Type => Int = {
+    case TypePlaceholder => 5
+    case TypeIdentifier(n) => 7 * n.hashCode()
+    case FunType(inT, outT) => 11 * inT.hashCode() + 13 * outT.hashCode() + 1
+    case DepFunType(_, t) => 17 * t.hashCode()
+    case dataType: DataType => 19 * dataType.hashCode()
```
Bastacyclop, line 74: this does not seem to respect alpha-equivalence for TypeIdentifier and other identifiers found in DataType.

umazalakain:
@Bastacyclop I made some tests and the only place in elevate where `=~~=` does not work and `=~=` has to be used is in `rise.elevate.rules.traversal`

src/main/scala/rise/elevate/rules/traversal.scala
```diff
@@ -58,7 +58,7 @@ object traversal {
     extends Strategy[Rise] {
 
     def apply(e: Rise): RewriteResult[Rise] = e match {
-      case ap @ App(f, x) if f == p => s(x).mapSuccess(App(f, _)(ap.t))
+      case ap @ App(f, x) if f =~= p => s(x).mapSuccess(App(f, _)(ap.t))
```
Bastacyclop, line 61: do we traverse untyped expressions sometimes?

src/test/scala/rise/elevate/algorithmic.scala
```diff
@@ -135,7 +135,7 @@ class algorithmic extends test_util.Tests {
     val typedGold = goldMKNAlternative.toExpr
     val loopMKN = (topDown(liftReduce) `;` DFNF `;` topDown(removeTransposePair)).apply(mm).get
 
-    assert(loopMKN == typedGold)
+    assert(loopMKN =~= typedGold)
```
Bastacyclop, line 138: `=~~=`

src/test/scala/rise/elevate/algorithmic.scala
```diff
@@ -238,7 +238,7 @@ class algorithmic extends test_util.Tests {
 
     val loopKMN = topDown(liftReduce).apply(mmMKN.toExpr).get
 
-    assert(goldKMNAlternative2.toExpr == loopKMN)
+    assert(goldKMNAlternative2.toExpr =~= loopKMN)
```
Bastacyclop, line 241: `=~~=`?

src/test/scala/rise/elevate/tiling.scala
```diff
@@ -325,10 +322,10 @@ class tiling extends test_util.Tests {
...
 
     val gold2 = DFNF(λ(i => λ(f => (J o **(f) o S) $ i))).get
-    assert(makeClosed((DFNF `;` DFNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get) == makeClosed(gold2))
-    assert(makeClosed((DFNF `;` DFNF `;` DFNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get) == makeClosed(gold2))
+    assert(makeClosed((DFNF `;` DFNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get) =~= makeClosed(gold2))
+    assert(makeClosed((DFNF `;` DFNF `;` DFNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get) =~= makeClosed(gold2))
 
     val gold3 = (DFNF `;` RNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get
-    assert(makeClosed((DFNF `;` RNF `;` DFNF `;` RNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get) == makeClosed(gold3))
+    assert(makeClosed((DFNF `;` RNF `;` DFNF `;` RNF)(λ(i => λ(f => (J o **(f) o S) $ i))).get) =~= makeClosed(gold3))
```
Bastacyclop, line 329: `=~~=`?

src/test/scala/rise/elevate/tiling.scala
```diff
@@ -34,8 +34,7 @@ class tiling extends test_util.Tests {
   def betaEtaEquals(a: Rise, b: Rise): Boolean = {
     val na = BENF(a).get
     val nb = BENF(b).get
-    val uab: Rise = toBeTyped(na) !: nb.t
-    makeClosed(uab) == makeClosed(nb)
+    makeClosed(toBeTyped(na)) =~= makeClosed(nb)
```
Bastacyclop, line 37: is there a reason to remove the type annotation `!: nb.t`?

src/test/scala/rise/elevate/util/package.scala
```diff
@@ -11,44 +11,11 @@ package object util {
 
   // Rise-related utils
 
-  def betaEtaEquals(a: Rise, b: Rise)(implicit ev: Traversable[Rise]): Boolean = BENF()(ev)(makeClosed(a)._1).get == BENF()(ev)(makeClosed(b)._1).get
+  def betaEtaEquals(a: Rise, b: Rise)(implicit ev: Traversable[Rise]): Boolean =
+    BENF()(ev)(makeClosed(a)).get =~= BENF()(ev)(makeClosed(b)).get
```
Bastacyclop, line 15: `=~~=`?

src/test/scala/rise/elevate/circularBuffering.scala
```diff
@@ -111,7 +111,7 @@ class circularBuffering extends test_util.Tests {
   private def openEquality(typedA: Rise, b: Rise): Boolean = {
     import rise.core.DSL._
     val typedB: Rise = toBeTyped(b) !: typedA.t
-    makeClosed(typedA)._1 == makeClosed(typedB)._1
+    makeClosed(typedA) =~= makeClosed(typedB)
```
Bastacyclop, line 114: `=~~=`?

src/main/scala/rise/elevate/rules/traversal.scala
umazalakain: That seems to be the case, changing it to `=~~=` breaks all of the elevate tests.

src/test/scala/rise/elevate/tiling.scala
umazalakain: It makes constraint solving fail, but I don't yet know why

umazalakain: So it appears to be something related to the type annotation making some implicit variables explicit, I need to investigate more

umazalakain: Fixed! Eta reduction wasn't operating correctly: it relies on the comparison of identifier names, without taking into account their types.

src/test/scala/rise/core/showScalaTest.scala
```diff
@@ -37,17 +37,11 @@ class showScalaTest extends test_util.Tests {
...
     val expr = toolbox.eval(toolbox.parse(untypedScala)).asInstanceOf[Expr]
     val typedExpr = toolbox.eval(toolbox.parse(typedScala)).asInstanceOf[Expr]
 
-    println(expr)
-    println(typedExpr)
-
-    assert(expr == dotElemWeights.toUntypedExpr)
-    assert(typedExpr == typedDotElemWeights)
+    assert(expr =~= dotElemWeights.toUntypedExpr)
+    assert(typedExpr =~= typedDotElemWeights)
```
Bastacyclop, line 45: would `=~~=` work here (second assert, not the first)?

src/main/scala/rise/elevate/rules/package.scala
```diff
@@ -40,7 +40,7 @@ package object rules {
 
   //TODO @rule
   def etaReduction()(implicit ev: Traversable[Rise]): Strategy[Rise] = {
-    case e@Lambda(x1, App(f, x2)) if x1 == x2 && !contains[Rise](x1).apply(f) => Success(f !: e.t)
+    case e@Lambda(x1, App(f, x2)) if x1 =~~= x2 && !contains[Rise](x1).apply(f) => Success(f !: e.t)
```
Bastacyclop, line 43: I do not understand, why would comparing identifiers by name break this rule? Do we have different types for the same identifier name? Would this not be a bug?

Bastacyclop (changes requested):
Looks good to me! Only a couple of minor comments.

src/main/scala/rise/core/equality.scala
```diff
@@ -0,0 +1,183 @@
...
+        case (_, TypePlaceholder) => true
+        case _ => typeAlphaEq.equiv[TypeKind](env)(a)(b)
+      }
+      case _ => typeAlphaEq.equiv(env)(a)(b)
+    }
+  }
+
+  object typeAlphaEq extends TypeEq {
+    /** Alpha equivalence on types.
+      * Short circuits in the event of syntactic equality.
```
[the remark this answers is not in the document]
umazalakain: Did some benchmarks on the test suite:
- `master`: 5min28sec
- `equality`: 3min59sec
- `equality` without the short-circuit: 4min9sec

It might be worth getting rid of the short-circuiting and go for the more straightforward behavior.

src/main/scala/rise/core/equality.scala
```diff
@@ -0,0 +1,183 @@
...
+  case class exprAlphaEq(typeEq : TypeEq) extends ExprEq {
+    /** Alpha equivalence on expressions.
+      * Parametrised by an equality relation on type.
+      * Kind equality is checked on dependent functions and pairs.
+      */
+    override val equiv: Env[Kind.Identifier] => Env[String] => Eq = typeEnv => exprEnv => a => b => {
+      val and = PatternMatching.matchWithDefault(b, false) // Make the match exhaustive
+      typeEq.equiv[TypeKind](typeEnv)(a.t)(b.t) && (a match {
+        case Identifier(na) => and { case Identifier(nb) => exprEnv.check(na, nb)}
+        case Literal(da) => and { case Literal(db) => da == db }
```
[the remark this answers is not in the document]
umazalakain: Aha! I presumed literals don't contain identifiers, fixed.

src/main/scala/rise/core/equality.scala
```diff
@@ -0,0 +1,183 @@
...
+        case NatToNatLambda(na, ba) => and { case NatToNatLambda(nb, bb) => equivNat(env.add(na, nb))(ba)(bb) }
+        case NatToDataLambda(na, ba) => and { case NatToDataLambda(nb, bb) => equiv[DataKind](env.add(na, nb))(ba)(bb) }
+        case NatCollectionFromArray(a) => and { case NatCollectionFromArray(b) => a == b } // FIXME: should use exprEq
+      })
+    }
+    val equivType: Env[Kind.Identifier] => Type => Type => Boolean = env => a => b => {
+      val and = PatternMatching.matchWithDefault(b, false)
+      a == b || (a match {
+        // Base cases
+        case TypePlaceholder => and { case TypePlaceholder => true }
```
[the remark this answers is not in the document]
umazalakain: I think we should keep it as is since `TypePlaceholder`s are `Type`s. Ideally this wouldn't be the case, we would keep `Option[Type]` where relevant and be able to talk of `Type`s without `TypePlaceholder`.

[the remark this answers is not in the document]
umazalakain: (Also, backwards compatibility: this was the old behavior as well.)

src/test/scala/rise/core/showScalaTest.scala
umazalakain: Doing `typedExpr.t =~~= typedDotElemWeights.t` works, but `typedExpr =~~= typedDotElemWeights` doesn't. One of the types within must have a placeholder that has escaped inference. Is it ok if I leave it as it is for now and investigate later?

src/test/scala/apps/separableConvolution2DEqsat.scala
```diff
@@ -29,25 +29,12 @@ class separableConvolution2DEqsat extends test_util.Tests {
...
-      case Literal(_: IndexData) => 93
-      case Literal(_: ArrayData) => 95
-      case Literal(_: PairData) => 97
-      case p: Primitive => p.getClass.hashCode()
+    override def hashCode(): Int = exprAlphaEq(typeErasure).hash(e)
+    override def equals(o : Any) : Boolean = o match {
+      case other : ExprWrapper => exprAlphaEq(typeUnificationAlphaEq).apply(this.e)(other.e)
+      case other : Expr => exprAlphaEq(typeUnificationAlphaEq).apply(this.e)(other)
+      case _ => false
     }
```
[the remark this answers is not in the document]
umazalakain: Yes, fixed

src/main/scala/rise/elevate/rules/package.scala
umazalakain: You are right! Using `==` didn't work for a while because `freeze` was broken, now that that is fixed `==` seems to work.

src/main/scala/rise/core/equality.scala
```diff
@@ -0,0 +1,183 @@
...
+    def equiv[K <: Kind] : Env[Kind.Identifier] => Eq[K]
+    def hash[K <: Kind] : K#T => Int
+  }
+
+  object typeErasure extends TypeEq {
+    override def hash[K <: Kind]: K#T => Int = _ => 0
+    override def equiv[K <: Kind]: Env[Kind.Identifier] => Eq[K] = _ => _ => _ => true
+  }
+
+  object typeUnificationAlphaEq extends TypeEq {
```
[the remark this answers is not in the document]
umazalakain: Done! While we are at it I switched `=~=` and `=~~=`: the idea is that the more tildes the less strict: `==` subset `=~=` subset `=~~=`.

umazalakain:
Weird, the `mmTensorCheck` passes locally for me, but doesn't pass GH's CI.

Bastacyclop (approved):

Bastacyclop (approved):

PR #114, opened 2021-02-24
Equality saturation experiments with naive data-structure

Bastacyclop:
A custom data-structure has also been investigated in https://github.com/rise-lang/shine/tree/eqsatCustom, but the next step is probably going to be using a proper e-graph inspired from https://github.com/egraphs-good/egg


michel-steuwer (approved):
👍

PR #116, opened 2021-03-05
OpenCL tests fail gracefully if the OpenCL library can't be loaded

michel-steuwer:
This PR makes all tests ineracting with the OpenCL executor fail gracefully if the executor library can not be loaded from disk (e.g. when it is not available on a specific development platform).

This PR guards against all code that failes do to a `UnsatisfiedLinkError` that is called from within a `test` function.

### DIFF
--- a/src/test/scala/test_util/package.scala
+++ b/src/test/scala/test_util/package.scala
@@ -1,5 +1,6 @@
 import opencl.executor.Executor
-import org.scalatest.BeforeAndAfter
+import org.scalactic.source.Position
+import org.scalatest.{BeforeAndAfter, Tag}
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.funsuite.AnyFunSuite
 import util.{AssertSame, Time, TimeSpan}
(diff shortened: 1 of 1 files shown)


Bastacyclop (commented):
@michel-steuwer That looks fine for `TestsWithExecutor`, but we also use `util.withExecutor` sometimes (e.g. for `shine.DPIA.Primitives.Scatter`), as well as `util.ExecuteOpenCL` when using host code generation.

michel-steuwer:
> @michel-steuwer That looks fine for `TestsWithExecutor`, but we also use `util.withExecutor` sometimes (e.g. for `shine.DPIA.Primitives.Scatter`), as well as `util.ExecuteOpenCL` when using host code generation.

I wasn't aware of this, but I have added support for this now.

Bastacyclop:
For `util.ExecuteOpenCL` maybe we can try running something like `clang -lOpenCL trivial.c`. If it fails, OpenCL is not installed? Maybe we actually then put the `isOpenCLAvailable` in a global variable and use that for all three cases? (`TestsWithExecutor`, `util.withExecutor`, and `util.ExecuteOpenCL`)

umazalakain:
What about exposing this opt out switch to the user? That way we keep things explicit and not engage in guess work.

Bastacyclop:
@umazalakain @michel-steuwer I like automatic support detection, maybe we can detect support by default while exposing override switches to bypass detection?

Bastacyclop:
So if `clang -lOpenCL test.c` works on my machine, OpenCL tests will be enabled by default. But I can still set `RISE_NO_OPENCL=1` or whatever variable to disable them.

umazalakain:
Sorry, somehow I missed these notifications. Automatic support detection + override switch sounds great!

umazalakain:
Maybe the only risk of having it be automatic is that the user might not be aware of whether they are running or not, mostly because currently the tests output *a lot* of information (I've been thinking of getting rid of all those print statements left in the tests, they make it more difficult to scroll through the output)

michel-steuwer:
We should aim for having a configuration with three states:
  `RISE_OPENCL=ENABLE`, `RISE_OPENCL=TRY`, `RISE_OPENCL=DISABLE`
  
We should also make this work for CUDA.

PR #117, opened 2021-03-08
CGO17 Benchmarks

Bastacyclop:
This PR makes sure that there is a benchmark function for each CGO17 benchmark, using the original input sizes and local/global sizes. I also updated the benchmarks status [on notion](https://www.notion.so/steuwer/Heuristic-Search-251f54f478ef4f31b5e72efac1a351d6) with various notes. Overall the main issues or points to keep in mind are:
- For the `convolution` benchmark, one implementation uses padding to avoid bank conflicts but this needs work to be properly ported over.
- For the `mm` benchmark, the ported NVIDIA implementation has extra copies which need more work to be eliminated.
- For the `nbody` and `gemv` benchmarks the performance looks different, it should be investigated
- Most benchmarks have no high-level expression to start rewriting from, we need to make these up
- The inputs used are random, in the original paper specific datasets are used
- I have not tried to run the benchmarks on hardware targets which would be close to the ones used in the paper. I simply ran then using POCL on my laptop CPU.

Otherwise I have noticed here and there that different loops are unrolled, different barriers inserted and other differences in the generated kernels which can affect performance. I think that a next step would also be to use host code generation for these benchmarks instead of the executor, which should ease integration with the upcoming auto-tuning framework and automatically inject local/global work sizes into the kernels as this is not done for every benchmark yet.

@michel-steuwer @johanneslenfers


src/main/scala/apps/convolution.scala
```diff
@@ -14,13 +14,37 @@ import shine.OpenCL.KernelExecutor._
...
+    o padClamp2D(0, 0, 8, 8) $ matrix
+  ))
+
+  // FIXME: could not find original Lift expression, this is made up
+  val blurYHL: ToBeTyped[Expr] = depFun((n: Nat) => fun(
+    (n `.` n `.` f32) ->: (17 `.` f32) ->: (n `.` n `.` f32)
+  )((matrix, weights) =>
+    map(map(transpose >> dotElemWeightsHL(weights)))
+    o slide2D(17, 1, 1, 1)
+    o padClamp2D(8, 8, 0, 0) $ matrix
```
michel-steuwer, line 45: I know we don't have any guidance on this, but the change in this file is introducing two styles:
 - the piping `x |> f |> g` style, vs.
 - the traditional functional `f o g $ x` style.
 
  Should we aim to commit to a preferred style?
 
 I personally much prefer the piping style because reads, left-to-right, top-to-bottom which I believe is much simpler.

src/main/scala/apps/nearestNeighbour.scala
```diff
@@ -13,6 +14,13 @@ object nearestNeighbour {
...
   )
 
+  // FIXME: could not find original Lift expression, this is made up
+  val highLevel: ToBeTyped[Expr] = depFun((n: Nat) => fun(
+    (n`.`(f32 x f32)) ->: f32 ->: f32 ->: (n`.`f32)
+  )((locations, lat, lng) =>
+    locations |> map(fun(loc => distance(loc)(lat)(lng)))
+  ))
+
   val nn: ToBeTyped[Expr] = depFun((n: Nat) => fun(
```
michel-steuwer, line 24: For (some) of the other apps you went for `nameHL` and `nameLL`, here you have `highLevel` vs. `name`.
Could we have a unified naming scheme?

michel-steuwer (commented):
Thanks @Bastacyclop.
:+1: which a couple of minor points to address.

src/main/scala/apps/convolution.scala
Bastacyclop: I also prefer the piping style, the question is if we want to keep the style that was in original Lift or not (it makes expression comparison easier when looking for potential differences / mistakes).

src/main/scala/apps/nearestNeighbour.scala
Bastacyclop: Yes we can do that, do you have a preference between the two? I think that `highLevel` can be too vague sometimes (i.e. if we have multiple highLevel expression like for mriQ benchmarks). We could also go for a combination of both like `nameHighLevel` and `nameLowLevel` (or `nameNvidia`, basically any implementation qualifier).

src/main/scala/apps/convolution.scala
michel-steuwer: I see. I would say: we should prefer the piping style by default and we can use the other one if we have a good reason for that. I would also say, that it would make sense to add a brief comment on why that style is used so that we don't refactor it by accident away.

src/main/scala/apps/nearestNeighbour.scala
michel-steuwer: `nameHighLevel` vs `nameLowLevel` sound good to me. The name could sometimes include a version, so `nameVersionHighLevel` (as in `mmNvidiaHighLevel`).

src/main/scala/apps/convolution.scala
Bastacyclop: I changed one or two expression to the piping style. Otherwise I am fine if the other expressions get refactored later on, this was just easier short-term. Maybe we can merge as-is and people are free to refactor to the piping style if and when they desire?

src/main/scala/apps/nearestNeighbour.scala
Bastacyclop: I tried to make the names more consistent, let me know if that is good enough.

michel-steuwer (approved):
👍

PR #118, opened 2021-03-10
base to scanline using mapLastFission instead of mapFirstFission

Bastacyclop:
This is good to have since `mapLastFission` is an almost syntactic rewrite while `mapFirstFission` is a complicated scala procedure (which should ideally be an Elevate strategy).


michel-steuwer (approved):
👍

PR #120, opened 2021-03-13
Remove unfreeze: use `++` on `explicit -> implicit` substitutitutions

umazalakain:


src/main/scala/rise/core/DSL/infer.scala
```diff
@@ -70,34 +75,24 @@ object infer {
...
-        ts.view.mapValues(t => ftvSubs(t)).toMap,
-        ns.view.mapValues(n => ftvSubs(n)).toMap,
-        as.view.mapValues(a => ftvSubs(a)).toMap,
-        n2ds.view.mapValues(n2d => ftvSubs(n2d)).toMap,
-        n2ns.view.mapValues(n2n => ftvSubs(n2n)).toMap,
-        natColls.view.mapValues(ftvSubs(_)).toMap
-      )
+  private def explToImpl[K <: Kind.Identifier] : K => Map[K, K] = {
+    case i : Kind.Explicitness => Map(i.asExplicit.asInstanceOf[K] -> i.asImplicit.asInstanceOf[K])
+    case i => Map(i -> i)
```
Bastacyclop, line 80: should that happen?

umazalakain: Good catch, we can even use types to rule it out ;)

Bastacyclop (approved):

PR #123, opened 2021-03-17
Porting the egg library

Bastacyclop:
This PR ports the [egg library](https://github.com/egraphs-good/egg) from Rust to Scala, while specialising the code for use with Rise. It also ports my [prototype combining egg and Rise](https://github.com/Bastacyclop/egg-rise).


Bastacyclop:
@michel-steuwer My prototype is now fully ported, some notes:
- The `egg` library is heavily documented, but I did not copy over all of the comments into this port, maybe I should. Or at least include a couple of links to `egg` docs.
- Performance on the biggest example is 10x slower (1s instead of 0.1s with `egg`). Part of this is due to scala or related implementation differences, and part of this is because one or two optimizations are not ported yet:
  - `ematching.forEachMatchingNode` has an optimized implementation when many equivalent nodes are in the e-class, I have not ported this yet
  - `egg` internalizes Strings (e.g. for pattern variable names), I am not sure what kind of performance impact this has
  - in `EGraph.rebuildClasses` the original library uses in-place sorting to remove two types of duplicates, this might have an important performance impact to investigate
- Rise expressions with De-Bruijn indices are implemented in `rise.eqsat.Expr` by re-using the generic `rise.eqsat.Node`, another option would be to extract this code outside of the `eqsat` package
- Types are currently erased and should be added to the implementation, maybe inside e-classes, since all equivalent nodes will share the same type? We could also think about hash-consing types to enable more structural sharing and faster equality checks.
- `DepApp`, `DepLambda` and `Nat` are not supported yet.
- Rewrite rules are duplicated in this new system (Elevate rewrite rules cannot be used), it would be good to factorise their definition as much as possible. The first step could be to abstract over DeBruijn indices vs named variables. DeBruijn index shifting should ideally be automatically infered.

michel-steuwer (approved):
Overall happy with the code.

A lot of new concepts and names are introduced here. Some high-level comments at the top of files or type definitions would be helpful. Maybe also just pointing to the egg rust documentation that explains these concepts in more depth?

michel-steuwer (approved):
That looks great. Thanks @Bastacyclop

Issue #106, opened 2021-02-11
Refactor OpenCL and CUDA backends


Bastacyclop:
After #100, the code should be refactored to avoid code duplication between CUDA and OpenCL.
Some notes about that:
- both backends implement `HoistMemoryAllocations` and `AdaptKernelBody` passes, we should look into merging them.
- there is `InjectWorkItemSizes` for OpenCL and `InjectThreadSizes` for CUDA, they basically do the same thing.
- CUDA has `mapBlock`/`mapThreads`/`mapGlobal`, OpenCL has `mapWorkgroup`/`mapLocal`/`mapGlobal`, we can have a common abstraction.  `mapWarp`/`mapLane` are also available in CUDA, in the future it might also make sense to have these for OpenCL (e.g. to use AMD wavefronts), but this is a device-dependent feature (it makes no sense on a CPU device). This also impacts `ParallelismLevel` and imperative `parFor`s.
- OpenCL has an `InsertMemoryBarriers` pass, something similar has to be done for CUDA and the code could be shared

PR #124, opened 2021-03-17
Benchmarking with host code generation

Bastacyclop:
This PR uses host code generation for one of the CGO17 benchmarks instead of the kernel executor.
It also adds support for host code generation with local memory. Local memory example:
```scala
val e = depFun((n: Nat) => fun((n`.`i32) ->: (n`.`i32))(in =>
      oclRun(LocalSize(16), GlobalSize(n))(
        in |> split(16) |> mapWorkGroup(0)(
          mapLocal(0)(add(li32(1))) >>
          toLocal >>
          mapLocal(0)(add(li32(2)))
        ) |> join
      )
    ))
```
Generates:
```c
// ...
    DeviceBuffer b0 = deviceBufferSync(ctx, moutput, (n1 * sizeof(int32_t)), DEVICE_WRITE);
    DeviceBuffer b2 = deviceBufferSync(ctx, me2, (n1 * sizeof(int32_t)), DEVICE_READ);
    const size_t global_size[3] = ((const size_t[3]){n1, 1, 1});
    const size_t local_size[3] = ((const size_t[3]){16, 1, 1});
    const KernelArg args[4] = ((const KernelArg[4]){KARG(b0), KARG(n1), KARG(b2), LARG((16 * sizeof(int32_t)))});
    launchKernel(ctx, k0, global_size, local_size, 4, args);
// ...
```


Bastacyclop:
For convolution we can compare the runtime with executor:
```
runtime over 6 runs for small
dpia X: 57.979822999999996 ms [55.026652 ms - 61.99596 ms]
dpia Y: 68.243513 ms [58.024927 ms - 70.669292 ms]
runtime over 3 runs for large
dpia X: 204.721422 ms [194.64940199999998 ms - 226.122585 ms]
dpia Y: 285.328533 ms [271.603857 ms - 287.425173 ms]
```
To the runtime with host code:
```
runtime over 6 runs for N=4096
dpia X: 0.128806 s [0.11852 s - 0.137337 s]
dpia Y: 0.148401 s [0.119374 s - 0.968686 s]
runtime over 3 runs for N=8192
dpia X: 0.397979 s [0.368228 s - 0.482509 s]
dpia Y: 0.435422 s [0.424763 s - 0.52969 s]
```
Which is clearly not satisfying. The first thing we should do to reduce host code overhead is to cache the kernel creation, since these timing include compiling the OpenCL kernels on every run. Alternatively we might want to use OpenCL events to isolate the kernel runtime like is done with the executor.

Bastacyclop:
Here is the fully assembled benchmark code for one of the convolution benchmark, if you want to inspect:
```c
const char k0_source[] =
""
""
""
"__kernel __attribute__ ((reqd_work_group_size(16, 4, 1)))"
"void k0(global float* restrict output, const global float* restrict e2, const global float* restrict e3, local float* restrict x1360){"
// ...
"}"
"";

#define loadKernel(ctx, id)\
  loadKernelFromSource(ctx, #id, id##_source, sizeof(id##_source) - 1)

#include "runtime.h"

void fun(Context ctx, Buffer moutput, Buffer me2, Buffer me3){
  {
    Kernel k0 = loadKernel(ctx, k0);
    DeviceBuffer b0 = deviceBufferSync(ctx, moutput, (4096 * (4096 * sizeof(float))), DEVICE_WRITE);
    DeviceBuffer b1 = deviceBufferSync(ctx, me2, (4096 * (4096 * sizeof(float))), DEVICE_READ);
    DeviceBuffer b2 = deviceBufferSync(ctx, me3, (17 * sizeof(float)), DEVICE_READ);
    const size_t global_size[3] = ((const size_t[3]){512, 4096, 1});
    const size_t local_size[3] = ((const size_t[3]){16, 4, 1});
    const KernelArg args[4] = ((const KernelArg[4]){KARG(b0), KARG(b1), KARG(b2), LARG((4 * (144 * sizeof(float))))});
    launchKernel(ctx, k0, global_size, local_size, 4, args);
    destroyKernel(ctx, k0);
  }
  
}




int main() {
  Context ctx = createDefaultContext();
  for (int sample = 0; sample < 6; sample++) {
    
if (sample == 0) srand(time(NULL));
Context ctx = createDefaultContext();
Buffer matrix = createBuffer(ctx, 4096 * 4096 * sizeof(float), HOST_READ | HOST_WRITE | DEVICE_READ);
Buffer weights = createBuffer(ctx, 17 * sizeof(float), HOST_READ | HOST_WRITE | DEVICE_READ);
Buffer output = createBuffer(ctx, 4096 * 4096 * sizeof(float), HOST_READ | HOST_WRITE | DEVICE_WRITE);

float* m = hostBufferSync(ctx, matrix, 4096 * 4096 * sizeof(float), HOST_WRITE);
for (int i = 0; i < 4096 * 4096; i++) {
  m[i] = (float)(rand())/(float)(RAND_MAX) * 10.0f;
}

float* w = hostBufferSync(ctx, weights, 17 * sizeof(float), HOST_WRITE);
for (int i = 0; i < 17; i++) {
  w[i] = (float)(rand())/(float)(RAND_MAX);
}

    Instant start = now();
    
fun(ctx, output, matrix, weights);
waitFinished(ctx);

    Instant stop = now();
    printf("sample: %lf\n", elapsedSeconds(start, stop));
    
// TODO: could check output here

destroyBuffer(ctx, matrix);
destroyBuffer(ctx, weights);
destroyBuffer(ctx, output);
destroyContext(ctx);

  }
}
```

src/main/scala/util/gen.scala
```diff
@@ -275,18 +277,19 @@ object gen {
...
+          hostFunDefToHostPart)
     }
 
-    private def hostFunDefToHostPart(gen: HostCodeGenerator =
-                                     shine.OpenCL.Compilation.HostCodeGenerator()
-                                    ): FunDef => CModule =
-      HostCodeModuleGenerator.funDefToModule(gen)
+    private val hostFunDefToHostPart:
+      ((FunDef, Seq[KernelModule])) => (CModule, Seq[KernelModule]) = { case (hm, kms) =>
+      ???
```
michel-steuwer, line 286: What is going on here?

michel-steuwer (changes requested):
Looks good to me. The `???` should either be explained in a comment or replaced by an implementation.

src/main/scala/util/gen.scala
Bastacyclop: The OpenCL host code generator now needs access to the dependent OpenCL kernel modules, so we cannot use the OpenCL host code generator for CUDA kernel modules without generalizing the implementation. We should probably have a generic host code generator and generic kernel modules as part of #106 . I can add a comment in the code.

michel-steuwer: Yes sounds good. Happy to have this merged then.

michel-steuwer (approved):
👍

PR #125, opened 2021-03-18
[Autotuning] Constraint checking and parameter substitution

Bastacyclop:


src/test/scala/rise/autotuning.scala
```diff
@@ -40,10 +40,38 @@ class autotuning extends test_util.Tests {
...
+    )
+    assert(!autotune.checkConstraints(constraints, badParameters2))
+
+    /* FIXME: there is no `n >= tile` constraint collected
+    val badParameters3 = Map(
+      NatIdentifier("vec", isExplicit = true) -> (8: Nat),
+      NatIdentifier("tile", isExplicit = true) -> (64: Nat)
+    )
+    assert(!autotune.checkConstraints(constraints, badParameters3))
+    */
```
johanneslenfers, line 67: we could add a tuning parameter `n` for the input size and give it a fixed value. Then we could use it in the constraints.

johanneslenfers (approved):
looks good to me

src/test/scala/rise/autotuning.scala
Bastacyclop: We can already talk about `n` inside the constraints, and there are already constraints involving `n`. The issue is that the constraint collection does not collect any constraint of the form `n >= tile` although it is required for validity.

PR #127, opened 2021-03-22
RISE Macro replacement

michel-steuwer:
Scala 3 does not support macros in the same way as Scala 2 does.

Some form of macros get much nicer (see e.g. https://github.com/elevate-lang/elevate/blob/scala-3/src/main/scala/elevate/core/macros/Macros.scala) but other forms of macros can not be implemented with Scala 3 macros at all.

Particularly in Scala 3 macros can not be used to generate new types. This is a direct consequence of the need for Scala 3 macros to be blackbox, i.e. `a macro expansion cannot influence the type of the expanded expression as seen from the typechecker` (https://www.scala-lang.org/blog/2018/04/30/in-a-nutshell.html).

This PR replaces in a first step the RISE `@primitive` macro with a code generator that is run at compile time via sbt and generates Scala files that are then compiled normally.

This approach has the big advantage that the IDE only sees the expanded code and does not have to guess how the expanded code and the macro definition relate to each other (this was a problem that we had with our use of macros before).

It might also open the door to generate `MLIR` implementations for our primitives.

This PR implements a parser of primitive declarations of the form:
```
def id: {t: data} -> t -> t;
def map: {n: nat} -> {dt1: data} -> {dt2: data} -> (dt1 -> dt2) -> n.dt1 -> n.dt2;
```
these are then expanded into classes that are equivalent to the objects and classes that the macro annotation generated before.

The primitives are declared in a `.rise` file and for each primitive, a new `.scala` file is generated in the same folder as the input with the name of the primitive.
This allows for easily defining primitives in scala in the old way for cases the parser can not handle yet, or where the type of the primitive is generated programmatically (like `makeArray`).


michel-steuwer:
@umazalakain this PR obviously also has a direct connection to #112

umazalakain:
Yes, this very much supersedes #112: the `.rise` files are so clean that we might as well present them as documentation. The only issue is with the primitives that cannot be autogenerated, but since there are so few maybe we could include them in the documentation by hand.

michel-steuwer:
@umazalakain and @Bastacyclop this is now ready for review (and hopefully to be merged).

meta/src/main/scala/meta/RisePrimitiveGenerator.scala
```diff
@@ -0,0 +1,268 @@
...
+      object ${Term.Name{name}} extends Builder {
+        private final case class Primitive()(override val t: Type = TypePlaceholder)
+          extends rise.core.Primitive
+        {
+          override val name: String = ${Lit.String(name)}
+          override def setType(ty: Type): Primitive = Primitive()(ty)
+          override def typeScheme: Type = ${generateTypeScheme(typeSignature, Map.empty)}
+
+          override def equals(obj: Any): Boolean = obj match {
+            case p: Primitive => p.t =~~= t
```
Bastacyclop, line 73: Should the default `equals` not be kept to be consistent with the `equals` of other language constructs?
https://github.com/rise-lang/shine/blob/5976011b902c1396acc145f2e66a4ff4a1fddfbd/src/main/scala/rise/core/Expr.scala#L28-L31

Bastacyclop (approved):
Looks great overall :+1:

meta/src/main/scala/meta/RisePrimitiveGenerator.scala
umazalakain: Alpha equivalence now considers two primitives to be the same if their names are the same, I think it's safe to do the same here.

meta/src/main/scala/meta/RisePrimitiveGenerator.scala
```diff
@@ -0,0 +1,268 @@
...
+        private final case class Primitive(..$params)(override val t: Type = TypePlaceholder)
+          extends rise.core.Primitive
+        {
+          override val name: String = ${Lit.String(name)}
+          override def setType(ty: Type): Primitive = Primitive(..$args)(ty)
+          override def typeScheme: Type = ${generateTypeScheme(typeSignature, Map.empty)}
+
+          override def equals(obj: Any): Boolean = obj match {
+            case p: Primitive =>
+              ${generateComparisonChain(args)} && p.t =~~= t
```
umazalakain, line 123: Same here, replacing `p.t =~~= t` with `p.name == name` should be ok

umazalakain (approved):
Looks great, particularly the kinding check

meta/src/main/scala/meta/RisePrimitiveGenerator.scala
michel-steuwer: I am doing exactly what the current macro in `master` is doing: https://github.com/rise-lang/shine/blob/master/macros/src/main/scala/rise/macros/Primitive.scala#L62

meta/src/main/scala/meta/RisePrimitiveGenerator.scala
michel-steuwer: As above: I am just doing what the current macro in `master is doing:
https://github.com/rise-lang/shine/blob/master/macros/src/main/scala/rise/macros/Primitive.scala#L101-#L106

meta/src/main/scala/meta/RisePrimitiveGenerator.scala
umazalakain: We could leave things as they are here and open a separate issue?

michel-steuwer: @umazalakain could you do this please and we merge this PR?

umazalakain: Done: #141

PR #129, opened 2021-03-26
Clean up type inference constraint collection

umazalakain:
Favors a more straightforward and explicit approach to the collection of type inference constraints where the identifier environment is immutable and only ever passed top-down, while the constraint collection is (also immutable) and passed bottom-up. It also seals the `Constraint` trait and gets rid of the unimplemented and non-generated `AddressSpaceConstraint`. Running `sbt test` on master and on this branch shows no performance impact.


michel-steuwer (approved):
👍

PR #130, opened 2021-03-26
Substitute stdout printing in tests with calls to log4j

umazalakain:
The test output is very verbose, to the point that it is sometimes difficult to scroll to the bits you are interested in. Simply removing prints is an option, but it could alter the semantics of the test, and there is some value in having them around, pin pointing what the interesting parts of the are.

Log4j seems to be the standard logging library in Scala, sbt uses it too. log4j-scala is a small wrapper around it that makes it somewhat easier to use in Scala. I have substituted the calls to print in the tests with calls to `logger.debug`. Log4j is configured via a configuration file in `src/main/resources/log4j2.xml`, which in our case defaults to the logging level `warn`. Ideally I would have this logging level set by sbt on test runs, but that seems surprisingly complicated: sbt passes configuration options to scalatest, which would then need to somehow pass these options to log4j. I couldn't find support for arbitrary options in scalatest, and as far as I am aware configuring logging levels in log4j can only be done from the configuration file.

We should maybe also decide if we would want something similar for `src/main`.


michel-steuwer (dismissed):
Very nice 👍

Bastacyclop:
@umazalakain so if a test fails and I want to see the logs, how would I do this?

michel-steuwer:
> @umazalakain so if a test fails and I want to see the logs, how would I do this?

I have made a minor change so that you can override the default log level (which is `WARN`) using the `RISE_LOG_LEVEL` environment variable:
```
RISE_LOG_LEVEL=DEBUG sbt test
```
spits out all message with log level `DEBUG` or above.

You can obviously also set the environment variable inside of IDEA when running a single test.

Bastacyclop (approved):
:+1:

umazalakain:
Nice, this makes it a lot more usable

PR #133, opened 2021-03-31
Avoid generating unnecessary constraints for lambdas and deplambdas

umazalakain:
Minor change that avoids generating constraints on fresh identifiers that are not mentioned anywhere else


src/main/scala/rise/core/DSL/infer.scala
```diff
@@ -136,9 +139,8 @@ object infer {
       val env1 : Map[String, Type] = env + (tx.name -> tx.t)
       val (te, cs, ftvE) = constrainTypes(env1)(e)
       val ft = FunType(tx.t, te.t)
-      val exprT = genType(expr)
-      val c = TypeConstraint(exprT, ft)
-      (Lambda(tx, te)(ft), cs :+ c, ftvE)
+      val cs1 = constrIfTyped(expr.t)(TypeConstraint(expr.t, ft))
+      (Lambda(tx, te)(ft), cs :++ cs1, ftvE)
```
Bastacyclop, line 143: minor style: we usually use `++` instead of `:++`, I think that they do the same.

Bastacyclop (commented):
@umazalakain what is the motivation for this change? does it make inference faster?

umazalakain:
@Bastacyclop It avoids adding (and solving) constraints that do not constraint type inference in any way whatsoever. The tests run in 4min5sec on this branch versus 4min49sec on master, though one of the `mmTensorCheck` tests fails on here, so it might be slightly misleading.

src/main/scala/rise/core/DSL/infer.scala
umazalakain: Fixed

michel-steuwer (approved):
Fine with me. @umazalakain you will have to resolve a merge conflict first.

michel-steuwer (approved):
👍

PR #135, opened 2021-04-01
Gather FTVs to preserve, hand them to constraint resolution

### LINKED ISSUE #128: Add support for directed type inference constraints
We currently use `freeze` and `unfreeze` to make type inference directed by converting implicit type variables into explicit type variables, solve constraints (which will substitute implicits with explicits but never the other way around), and convert back. This is very unintuitive and prone to errors. We could instead have explicit `Undirected : Type -> Type -> Constraint` and `Directed : Type -> Type -> Constraint`.

umazalakain:
Closes #128 

- Introduce a pre-inference pass `collectPreserve` that collects the type variables to preserve from type assertions and opaques. While doing so, it transforms all type assertions into type annotations. This makes it easier to eventually drop type assertions altogether (#142).
- Remove freeze/unfreeze, instead pass to constraint solving a list of type identifiers to preserve. The code style could maybe be improved here by making `preserve` global to all the functions called within `solveOne`.

### DIFF
--- a/src/main/scala/rise/core/DSL/infer.scala
+++ b/src/main/scala/rise/core/DSL/infer.scala
@@ -10,14 +10,24 @@ import rise.core.types._
 import scala.collection.mutable
 
 object infer {
+  case class V(sol: Solution) extends SolutionVisitor(sol) {
+    override def `type`[T <: Type] : T => Pure[T] = t => return_(sol(t).asInstanceOf[T])
+    override def expr: Expr => Pure[Expr] = {
+      case Opaque(e, t) => for {t1 <- `type`(t)} yield Opaque(e, t1)
+      case TypeAnnotation(e, t) => for {t1 <- `type`(t)} yield TypeAnnotation(e, t1)
+      case e => super.expr(e)
+    }
+  }
   private [DSL] def apply(e: Expr,
             printFlag: Flags.PrintTypesAndTypeHoles = Flags.PrintTypesAndTypeHoles.Off,
             explDep: Flags.ExplicitDependence = Flags.ExplicitDependence.Off): Expr = {
+    val (collected_e, ftvSubs) = collectAssertOpaque(e)
+    val frozen_e = V(freeze(ftvSubs)).expr(collected_e).unwrap
     // Constraints of the form `implicit type var == explicit type var` result in substitutions
     // `implicit type var -> explicit type var`. We (ab)use that fact to create directed constraints out of
     // type assertions and opaque types. To do so, we make the type identifiers on one side of the constraint explicit,
     // and we return a `ftvSubs` map that maps these explicit type identifiers back to implicit type identifiers.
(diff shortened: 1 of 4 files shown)


src/main/scala/rise/core/types/Constraints.scala
```diff
@@ -44,9 +44,12 @@ case class NatCollectionConstraint(a: NatCollection, b: NatCollection)
 }
 
 object Constraint {
-  def solve(cs: Seq[Constraint], trace: Seq[Constraint])
+  val canBeSubstituted : Seq[Kind.Identifier] => Kind.Identifier with Kind.Explicitness => Boolean =
+    preserve => i => !(preserve.contains(i) || i.isExplicit)
```
Bastacyclop, line 48: would it make sense to also collect explicit variabels into `preserve` to simplify this test?

Bastacyclop: it would then make sense to replace the `asImplicit`/`asExplicit` kind of transformations during type inference into `preserve` additions and substractions?
https://github.com/rise-lang/shine/pull/135/files#diff-b111bbb431e2b7263963fdc5d76146aef69e6a754dc981feb1d410e3feb28605L140-L147

src/main/scala/rise/core/DSL/infer.scala
```diff
@@ -123,31 +78,49 @@ object infer {
 
   private val genType : Expr => Type = e => if (e.t == TypePlaceholder) freshTypeIdentifier else e.t
 
-  private val constrainTypes : Map[String, Type] => Expr => (Expr, Seq[Constraint], Solution) = env => {
+  private val collectPreserve = new PureAccumulatorTraversal[Seq[Kind.Identifier]] {
+    override val accumulator = SeqMonoid
```
Bastacyclop, line 82: I think that this should really be a `HashMap` for performance reasons? Ideally indentifiers should also not be string-based in the future (https://github.com/rise-lang/shine/issues/137).

src/main/scala/rise/core/types/Constraints.scala
umazalakain: I think it depends on what we do with explicitness information: if we distinguish them by having two nodes `BoundedVar` and `GloballyFreeVar` (along the lines of what @michel-steuwer was suggesting), then constraint solving would still have to deal with them. @Bastacyclop should we leave this as it is here, and I open an issue for how to encode explicitness metadata and I mention it there?

src/main/scala/rise/core/DSL/infer.scala
umazalakain: I put them in a `HashSet` -- and yes, when identifiers become integer based so does this.

src/main/scala/rise/core/types/Constraints.scala
umazalakain: Opened an issue #151

Bastacyclop (approved):

PR #136, opened 2021-04-01
Add a monad on `Monad[(Monoid[F], S)]`

umazalakain:
A low importance PR, but it adds a monad on pairs where the first element is a monoid, so we don't need to keep surrounding traversals with mutable environments on which we accumulate stuff. Albeit a small benefit, it allow us to merge `getFTVs` and `getFTVsRec` into a single traversal definition. It can also be used on the DPIA side once I finish refactoring the traversals there too.

### DIFF
--- a/src/main/scala/rise/core/DSL/infer.scala
+++ b/src/main/scala/rise/core/DSL/infer.scala
@@ -28,13 +28,12 @@ object infer {
   }
 
   def printTypesAndTypeHoles(expr: Expr): Unit = {
-    val hasHoles = new PairMonoidTraversal[Boolean, Pure] {
-      override val fstMonoid = OrMonoid
-      override val wrapperMonad = PureMonad
+    val hasHoles = new PureAccumulatorTraversal[Boolean] {
+      override val accumulator = OrMonoid
       override def expr: Expr => Pair[Expr] = {
         case h@primitives.typeHole(msg) =>
           println(s"found type hole ${msg}: ${h.t}")
-          record(true)(h : Expr)
+          accumulate(true)(h : Expr)
         case p@primitives.printType(msg) =>
           println(s"$msg : ${p.t} (Rise level)")
           return_(p: Expr)
@@ -98,12 +97,11 @@ object infer {
     )
   }
 
-  val FTVGathering = new PairMonoidTraversal[Seq[Kind.Identifier], Pure] {
-    override val fstMonoid = SeqMonoid
-    override val wrapperMonad = PureMonad
+  val FTVGathering = new PureAccumulatorTraversal[Seq[Kind.Identifier]] {
+    override val accumulator = SeqMonoid
     override def typeIdentifier[I <: Kind.Identifier]: VarType => I => Pair[I] = _ => {
-      case i: Kind.Explicitness => record(if (!i.isExplicit) Seq(i) else Seq())(i.asInstanceOf[I])
-      case i => record(Seq(i))(i)
+      case i: Kind.Explicitness => accumulate(if (!i.isExplicit) Seq(i) else Seq())(i.asInstanceOf[I])
+      case i => accumulate(Seq(i))(i)
     }
     override def nat: Nat => Pair[Nat] = ae => {
       val ftvs = mutable.ListBuffer[Kind.Identifier]()
@@ -111,7 +109,7 @@ object infer {
         case i: NatIdentifier if !i.isExplicit => ftvs += i; i
         case n => n
       })
-      record(ftvs.toSeq)(r)
+      accumulate(ftvs.toSeq)(r)
     }
   }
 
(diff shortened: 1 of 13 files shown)


src/main/scala/rise/core/DSL/infer.scala
```diff
@@ -27,21 +28,20 @@ object infer {
...
         case p@primitives.printType(msg) =>
           println(s"$msg : ${p.t} (Rise level)")
-          return_(p : Expr)
+          return_(p: Expr)
         case e => super.expr(e)
       }
-    })
-    if (holeFound) {
+    }
+    if (T.expr(expr).unwrap._1) {
```
Bastacyclop, line 44: I find mutability to work quite well for such "global accumulation" patterns. The monoidal version is more noisy and the code is harder to read for me.

src/main/scala/rise/core/uniqueNames.scala
```diff
@@ -1,34 +1,28 @@
...
 object uniqueNames {
-  private case class CountingVisitor(
-    var values: Map[Identifier, Int],
-    var types: Map[Kind.Identifier, Int]
-  ) extends PureTraversal {
-    override def identifier[I <: Identifier]: VarType => I => Pure[I] = {
-      case Binding => i =>
-        values = values + (i -> (values.getOrElse(i, 0) + 1))
-        return_(i)
+  private val CountingVisitor = new PairMonoidTraversal[(Seq[Identifier], Seq[Kind.Identifier]), Pure] {
```
Bastacyclop, line 8: this is not a `CountingVisitor` anymore I would say?

umazalakain: Fixed!

src/main/scala/rise/core/DSL/infer.scala
umazalakain: The trouble with that approach is that you cannot easily reuse the traversal (e.g. to use it once on an expression and once on a type). I have slightly reduced the noise by getting rid of the `unwrap`s and adding a `record` function so it is slightly more semantically meaningful.

Bastacyclop: With the mutable approach, can you not just accept the mutated accumulator as a parameter to the traversal/visitor to make it reusable? And to avoid surrounding the traversal with the environment on use can you not have a default accumulator and return it? like:
```scala
def traverseAndAccumulate(e: Expr, acc: mutable.HashMap = HashMap()): mutable.HashMap = {
  traverse(...) // acc += stuff
  acc
}
```
Anyway, we can also go for the Monad way, I just find it less straight-forward.

Bastacyclop (dismissed):

src/main/scala/rise/core/traverse.scala
```diff
@@ -196,23 +183,23 @@ object traverse {
...
-
-  implicit object OptionMonad extends Monad[Option] {
-    def return_[T]: T => Option[T] = Some(_)
-    def bind[T, S]: Option[T] => (T => Option[S]) => Option[S] = v => v.flatMap
-  }
-
-  trait PureTraversal extends Traversal[Pure] { override def monad = PureMonad }
+  trait PureTraversal extends Traversal[Pure] {override def monad : PureMonad.type = PureMonad }
   trait PureExprTraversal extends PureTraversal with ExprTraversal[Pure]
+  trait PairMonoidTraversal[F,M[_]] extends Traversal[InMonad[M]#SetFst[F]#Type] {
```
michel-steuwer, line 188: Can we name this `StatefulTraversal` with `fstMonoid` being renamed to `state`?

Or maybe `AccumulatorTraversal` with `fstMonoid` -> `accumulator`?

michel-steuwer: Could we then also default the Monad to `Pure` (as done for the `PureTraversal`)?

src/main/scala/rise/core/DSL/infer.scala
```diff
@@ -27,21 +28,20 @@ object infer {
   }
 
   def printTypesAndTypeHoles(expr: Expr): Unit = {
-    // TODO: move holeFound state into the traverse
-    var holeFound = false
-    traverse(expr, new PureExprTraversal {
-      override def expr : Expr => Pure[Expr] = {
+    val hasHoles = new PairMonoidTraversal[Boolean, Pure] {
+      override val fstMonoid = OrMonoid
+      override val wrapperMonad = PureMonad
```
michel-steuwer, line 33: With my other comments this should look like:
```scala
val hasHoles = new AccumulatorTraversal[Boolean] {
  override val acc = OrMonoid
  ...
}
```

michel-steuwer (changes requested):
I am not confinced of the benefits vs. complexity ratio here.

I think you can make the interface nicer by having less of the Monoid / Monad language coming through. At which point I am happy to have it merged.

src/main/scala/rise/core/traverse.scala
umazalakain: Done, thank you!

umazalakain:
@michel-steuwer renamed a bunch, looks better now!

michel-steuwer (approved):
Much better :-)

PR #138, opened 2021-04-04
Preliminary attempt at bidirectional typechecking/inference

umazalakain:
The main difference wrt what we currently have is that instead of three distinguished phases where we 1) collect constraints 2) solve constraints 3) apply substitutions, we traverse the expression tree alternating between inferring and checking, only ever creating type variables where we need to (i.e. abstractions that do not have typing annotations), and the few created constraints are solved eagerly in place.

The benefits wrt what we currently have are 1) much fewer constraints are generated 2) the errors are a lot more localised (if a constraint doesn't hold we know exactly where in the expression tree this constraint was created).

### DIFF
--- a/src/main/scala/rise/core/bidirectional.scala
+++ b/src/main/scala/rise/core/bidirectional.scala
@@ -3,6 +3,9 @@ package rise.core
 import scala.collection.mutable.ListBuffer
 import rise.core.semantics._
 import rise.core.types._
+import util.PatternMatching
+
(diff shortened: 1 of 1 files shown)


Bastacyclop:
This might require more type annotations in the Rise expressions, since the flow of information becomes fixed, is this correct?

Bastacyclop:
Otherwise for point 2) I think that we could also extend the current constraint trace mechanism to include the original expression localisation.

PR #144, opened 2021-04-07
Define two primitives as equal if their Scala classes are equal

### LINKED ISSUE #141: Redefine equality of RISE primitives based on their names
Alpha equivalence judges two primitives equal iff they have the same name. The macros for primitive generation however overwrite the `.equals` of primitives and judge two primitives equal if their types are alpha equivalent. We should either remove the `.equals` overwrite and deal with the consequences, or make it compare names in line with what alpha equivalence does.

umazalakain:
Closes #141


Bastacyclop:
@umazalakain Can you not keep the default `equals` implementation to achieve the same result?
```
object mapGlobal {
  private final case class Primitive(dim: Int)(override val t: Type = TypePlaceholder) extends rise.core.Primitive {
    // ...
    override def equals(obj: Any): Boolean = obj match {
      case p: Primitive => // this would already be tested by the default implementation
        p.dim == dim && true && // this would already be tested by the default implementation
        p.getClass == getClass // always true since we know p: Primitive ?
      case _ =>
        false
    }
  }
  // ...
```

umazalakain:
@umazalakain I don't think so, since we need to ignore the type `t`. But you have a point in that matching against `Primitive` (the private class, not `rise.core.Primitive`) already checks that `p.getClass == getClass`.

Bastacyclop:
We do not do anything in particular to ignore the type `t` for other language constructs like `App` as far as I know. So either it is already ignored because it is part of the second initialization list, or we are not implementing the same behaviour for constructs like `App`?

umazalakain:
@Bastacyclop So, for some reason last time I tried that things broke in a weird way (broken pipes between processes) *shrug*
It seems to be working though.

src/main/scala/rise/core/equality.scala
```diff
@@ -155,7 +155,7 @@ object equality {
       typeEq.equiv[TypeKind](typeEnv)(a.t)(b.t) && (a match {
         case Identifier(na) => and { case Identifier(nb) => exprEnv.check(na, nb)}
         case Literal(da) => and { case Literal(db) => equivData(typeEnv)(da)(db) }
-        case a: Primitive => and { case b: Primitive => a.name == b.name }
+        case a: Primitive => and { case b: Primitive => a.getClass == b.getClass }
```
Bastacyclop, line 158: should it not be `a == b` now then?

Bastacyclop: otherwise if `a == b` does not work (if it also looks at the type) then instead of overriding `equals` we could have a separate `primitiveA.samePrimitive(primitiveB)` method?

umazalakain: Actually, since some primitives might contain further expressions, should we not have this `samePrimitive` method take the equality environment generated by `=~=` so that it can keep checking further?

Bastacyclop: Primitives should not contain further expressions, they are only parameterised by compile-time constants that are outside of the language itself.

umazalakain: What about assertions, annotations and opaque expressions? Do we just assume that alpha equivalence will only run on expressions which had type inference ran on them and that all opaque expressions are equal?

Bastacyclop: Do we actually still use type annotation primitives? This is what happens in the DSL:
```scala
  implicit class TypeAnnotation(t: Type) {
    def ::(e: Expr): Expr = e.setType(t)
    def `:`(e: Expr): Expr = e.setType(t)
  }
```
The primitive was used in a previous DSL which has been removed, so maybe we can remove the primitive too?
We already discussed that we should remove type assertions.
And finally for `Opaque` it really looks like it should not be a primitive, unless we can rewrite it into something like `def opaque: {ŧ: Type} -> t -> t` or `def opaque: {t1: Type} -> {t2: Type} -> t1 -> t2` if that is not too clunky.

michel-steuwer: The code for type annotations that is used with `ToBeTyped` is this here:
```
implicit class TypeAnnotationHelper(t: Type) {
    def ::[T <: Expr](e: ToBeTyped[T]): ToBeTyped[Expr] =
      e >>= (e => toBeTyped(TypeAnnotation(e, t)))
  }
```

Bastacyclop: Ah sorry about that, looks like I was looking at the `FSmooth` DSL code.

Bastacyclop: Then I think we can close https://github.com/rise-lang/shine/issues/153 and think about  https://github.com/rise-lang/shine/issues/152.

umazalakain: I think for the time being and until we pull them out of the set of primitives, I think it is warranted to have especial cases for these "primitives" in alpha equivalence?

Bastacyclop: Sounds reasonable short-term.

michel-steuwer (approved):
Looks all good to me. Thanks, @umazalakain and @Bastacyclop for the discussion and opening #152

PR #147, opened 2021-04-10
Factored out translations from primitives

michel-steuwer:


src/main/scala/shine/DPIA/Compilation/AcceptorTranslation.scala
```diff
@@ -0,0 +1,356 @@
...
+
+    case VectorFromScalar(n, dt, arg) =>
+      con(arg)(λ(expT(dt, read))(e =>
+        A :=|VectorType(n, dt)| VectorFromScalar(n, dt, e)))
+
+    case Zip(n, dt1, dt2, access, e1, e2) =>
+      acc(e1)(ZipAcc1(n, dt1, dt2, A)) `;`
+        acc(e2)(ZipAcc2(n, dt1, dt2, A))
+
+    // OpenMP
```
Bastacyclop, line 252: Can we have instead a `shine.OpenMP.Compilation.AcceptorTranslation` which calls `shine.DPIA.Compilation.AcceptorTranslation`? Same for OpenCL and CUDA.

michel-steuwer: I think this is a good idea. I won't have time to do this over the next couple of days though ...

Bastacyclop (approved):

PR #148, opened 2021-04-10
Refactored reorder primitive to use NatToNat lambdas rather than idx[n] -> idx[n] functions

michel-steuwer:
Changes the type of the `reorder` primitive from:
```scala
def reorder: {n: nat} -> {t: data} -> (idx[n] -> idx[n]) -> (idx[n] -> idx[n]) -> n.t -> n.t
```
to
```scala
def reorder: {t: data} -> (n: nat) -> (idxF: nat2nat) -> (idxFinv: nat2nat) -> n.t -> n.t
```

This is how we should have implemented `reorder` from the start. But I believe initially we didn't had the `NatToNat` lambdas.

Before we wrote the reindexing functions like this:
```scala
val f = impl { n: Nat =>
	fun(IndexType(n))(i =>
		natAsIndex(n)(
		  (indexAsNat(i) / (n /^ s)) +
		    (l(s) * (indexAsNat(i) % (n /^ s)))
		)
	)
}
reorder(f)(f)
```

We had to convert awkwardly between `nat`s and `idx[n]`.
During code generation, we then used `evalIndexExp` to eventually turn the phrase with type `expr[idx[n], read]` into a `nat`.

This PR avoids these conventions.
Now we write the reindexing function like this:
```scala
impl { n: Nat =>
	def f = n2nFun(n)(i => (i / (n /^ s)) + (s * (i % (n /^ s))))
	reorder(n)(f)(f)
}
```

The new type of `reorder` does not express that the nat that is reindexed is smaller than `n`. I hope that we will be able to re-establish this in the type once we have a `Nat` language with the range information in the type.

This PR helps also to remove the partially implemented `eval` function (as the old `reorder` primitive requires the `evalIndexExp` functionality).

### DIFF
--- a/src/main/scala/rise/core/types/Solution.scala
+++ b/src/main/scala/rise/core/types/Solution.scala
@@ -1,7 +1,7 @@
 package rise.core.types
 
 import util.monads._
-import arithexpr.arithmetic.BoolExpr
+import arithexpr.arithmetic.{BoolExpr, NamedVar}
 import rise.core.traverse._
 import rise.core.{Expr, substitute}
 
@@ -104,6 +104,7 @@ case class Solution(ts: Map[Type, Type],
       case lambda@NatToNatLambda(x, body) =>
         val xSub = apply(x) match {
           case n: NatIdentifier => n
+          case n: arithexpr.arithmetic.NamedVar => NatIdentifier(n.name, n.range, isExplicit = true)
           case other => throw NonIdentifierInBinderException(lambda, other)
         }
         NatToNatLambda(xSub, apply(body))
--- a/src/main/scala/shine/DPIA/Types/NatToNat.scala
+++ b/src/main/scala/shine/DPIA/Types/NatToNat.scala
@@ -34,12 +34,12 @@ object NatToNatLambda {
   }
 
   def apply(upperBound:Nat, id: NatIdentifier, body:Nat):NatToNatLambda = {
-    val x = NamedVar(freshName("n"), RangeAdd(0, upperBound, 1))
+    val x = NatIdentifier(freshName("n"), RangeAdd(0, upperBound, 1))
     NatToNatLambda(x, x => ArithExpr.substitute(body, Map((id, x))))
   }
 
   def apply(range: Range, id: NatIdentifier, body: Nat): NatToNatLambda = {
-    val x = NamedVar(freshName("n"), range)
+    val x = NatIdentifier(freshName("n"), range)
     NatToNatLambda(x, x => ArithExpr.substitute(body, Map((id, x))))
   }
 }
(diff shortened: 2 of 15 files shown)


fedepiz (approved):
Looks good to me! 

Although I do want to point out the obvious just in case,  that is that `idx[n] -> idx[n]` is more stringent than `nat -> nat`.  I guess something else that might be addressed in the "typed nats" rework?

michel-steuwer:
> Looks good to me!
> 
> Although I do want to point out the obvious just in case, that is that `idx[n] -> idx[n]` is more stringent than `nat -> nat`. I guess something else that might be addressed in the "typed nats" rework?

Yes exactly. As I already pointed out in the message above:
> The new type of reorder does not express that the nat that is reindexed is smaller than n. I hope that we will be able to re-establish this in the type once we have a Nat language with the range information in the type.

src/main/scala/rise/core/types/Solution.scala
```diff
@@ -104,6 +104,7 @@ case class Solution(ts: Map[Type, Type],
       case lambda@NatToNatLambda(x, body) =>
         val xSub = apply(x) match {
           case n: NatIdentifier => n
+          case n: arithexpr.arithmetic.NamedVar => NatIdentifier(n.name, n.range, isExplicit = true)
```
Bastacyclop, line 107: when does this happen? that might indicate a bug elsewhere, since the `NatIdentifier` information should not be lost.

src/main/scala/shine/DPIA/Types/NatToNat.scala
```diff
@@ -37,6 +37,11 @@ object NatToNatLambda {
     val x = NamedVar(freshName("n"), RangeAdd(0, upperBound, 1))
     NatToNatLambda(x, x => ArithExpr.substitute(body, Map((id, x))))
   }
+
+  def apply(range: Range, id: NatIdentifier, body: Nat): NatToNatLambda = {
+    val x = NamedVar(freshName("n"), range)
```
Bastacyclop, line 42: should these not be `NatIdentifier`s?

Bastacyclop (commented):
This looks fine to me as long as (1) NatToNat lambdas are working well and here to stay (2) they will be extended to include range information but that is already your plan.

src/main/scala/rise/core/types/Solution.scala
michel-steuwer: This addition was definitely required to get the code to work. I would suggest that this is an orthogonal issue top this PR, as I basically copied this solution from the handling of `NatToData` that matches for `NamedVar` as well.

src/main/scala/shine/DPIA/Types/NatToNat.scala
michel-steuwer: I have fixed this.

src/main/scala/rise/core/types/Solution.scala
michel-steuwer: Maybe you can create an issue for this and we merge this PR?

Bastacyclop (approved):

PR #156, opened 2021-04-13
Sobel filter example app for Gap8 processor

bpervan:
This is the initial version of the Sobel filter application targeting the Gap8 processor. Further improvements will be done as the implementation of the Gap8-specific code generator progresses.


michel-steuwer (approved):
👍

Issue #152, opened 2021-04-12
Should assertions, annotations and opaque be primitives?


umazalakain:
`TypeAssertion`, `TypeAnnotation` and `Opaque` are the only primitives that contain expressions. They often need to be treated specially by traversals (e.g. if you want to define alpha equivalence on expressions with annotations you better step into the annotations too). Would it be sensible to make them first class members of the AST? If #134 goes through we could potentially only ever have them in the untyped fragment of the language.

Bastacyclop:
I think that it makes sense for assertions, annotations and opaques to be proper language constructs because they can contain other language constructs: they are not leaf nodes like primitives should be in my opinion.

Issue #134, opened 2021-04-01
Distinguish between untyped and typed expressions at the type level


umazalakain:
Currently, `Expr`s have `Type`s, and `Type`s can be untyped by being `TypePlaceholder`s. Then we have `ToBeTyped`, which we use to erase all `Type`s with `TypePlaceholder`s and drive type inference. Places that assume an `Expr` is fully typed (i.e. has no `TypePlaceholder`s) cannot express this through types.

How desirable would it be to take `TypePlaceholder` out of `Type`, turn `Expr` into `Expr[T]` where for `e : Expr[T]` we have `e.t : T`, and then handle `Expr[Unit]` for untyped expressions and `Expr[Type]` for fully typed ones?

Bastacyclop:
@umazalakain @michel-steuwer I started ordering enhancement issues and their priority according to my opinion in: https://github.com/rise-lang/shine/projects/1
Maybe we can keep discussing priorities and maintaining this list accordingly?

umazalakain:
@Bastacyclop @michel-steuwer Sounds great!

michel-steuwer:
I think that I am generally fine with this change.
I don't think that we had significant problems with `TypePlaceholders` so far, but I think it would be nice to get rid of them as they don't exist in our formal system.

I assume that this means that primitives will also need to be polymorph then?
What are the consequences of this for the code we write/generate for them?

We could introduce type aliases like:
```
type UntypedExpr = Expr[Unit]
type TypedExpr = Expr[Type]
```

Maybe we could even get the naming in a way so that `Expr` refers to the typed expression?
(i.e. `type Expr = NEW_NAME[Type]` with a different name for the polymorphic )

Keeping the name `Expr` for typed expressions would mean the least disruption for the wider codebase.

Overall, I don't think this has particularly high priority, but could be a nice addition.

fedepiz:
This is an "SML-modules" take on the same problem

```
trait Primitive
  case object DoThings extends Primitive
  // ...

  sealed trait Type
  final case object IntType extends Type



  // The definition of an AST, with it's own abstract Type type member
  trait Ast {
    type TType

    sealed abstract class Expr {
      val typ:TType
    }

    final case class Prim(p:Primitive, typ: TType) extends Expr
    final case class Apply(a: Expr, b: Expr, typ: TType) extends Expr
    // etc etc
  }

  // Untyped instantiation of the Ast
  object UntypedAst extends Ast {
    type TType = Unit
  }


[...]

PR #157, opened 2021-04-14
Primitives should not contain subexpressions

umazalakain:


src/test/scala/apps/separableConvolution2DNaiveEqsat.scala
```diff
@@ -63,6 +63,9 @@ class separableConvolution2DNaiveEqsat extends test_util.Tests {
       case DepApp(f, x) => everywhere(s)(f).map(DepApp(_, x)(p.t))
       case Literal(_) => Nil
       case _: ForeignFunction => Nil
+      case _: TypeAnnotation => Nil
+      case _: TypeAssertion => Nil
+      case _: Opaque => Nil
```
Bastacyclop, line 68: these should throw a runtime error?

src/main/scala/shine/DPIA/InferAccessAnnotation.scala
```diff
@@ -133,6 +132,9 @@ private class InferAccessAnnotation {
         inferDepLambda(depL, ctx, isKernelParamFun)
       case depA: r.DepApp[_] =>
         inferDepApp(depA, ctx, addsKernelParam(e, isKernelParamFun))
+      case TypeAnnotation(e, _) => inferPhraseTypes(e, ctx, isKernelParamFun)
+      case TypeAssertion(e, _) => inferPhraseTypes(e, ctx, isKernelParamFun)
+      case Opaque(e, _) => inferPhraseTypes(e, ctx, isKernelParamFun)
```
Bastacyclop, line 137: these should throw an exception?

src/main/scala/rise/elevate/strategies/lowering.scala
```diff
@@ -40,6 +40,9 @@ object lowering {
       case _: Identifier      => Failure(extract(what))
       case _: Literal         => Failure(extract(what))
       case _: ForeignFunction => Failure(extract(what))
+      case _: TypeAnnotation  => Failure(extract(what))
+      case _: TypeAssertion   => Failure(extract(what))
+      case _: Opaque          => Failure(extract(what))
```
Bastacyclop, line 45: these should throw an exception?

src/main/scala/rise/elevate/rules/traversal.scala
```diff
@@ -189,6 +189,9 @@ object traversal {
               Some(s(f).mapSuccess(DepApp[NatToDataKind](_, n2d)(da.t)))
           }
           case Literal(_) => None
+          case TypeAssertion(_, _) => None
+          case TypeAnnotation(_, _) => None
+          case Opaque(_, _) => None
```
Bastacyclop, line 194: exception?

src/main/scala/rise/core/equality.scala
```diff
@@ -198,7 +205,10 @@ object equality {
       case Literal(_: IndexData) => 93
       case Literal(_: ArrayData) => 95
       case Literal(_: PairData) => 97
-      case p: Primitive => p.name.hashCode() + typeEq.hash[TypeKind](p.t)
+      case Opaque(e, t) => 101*hash(e) + 103*typeEq.hash[TypeKind](t)
+      case TypeAnnotation(e, t) => 107*hash(e) + 109*typeEq.hash[TypeKind](t)
+      case TypeAssertion(e, t) => 113*hash(e) + 127*typeEq.hash[TypeKind](t)
+      case p: Primitive => 131*p.name.hashCode() + 137*typeEq.hash[TypeKind](p.t)
```
Bastacyclop, line 211: can we use `p.hashCode()`?

Bastacyclop (changes requested):

src/main/scala/rise/core/equality.scala
umazalakain: No, since that would take the type into account.

src/test/scala/apps/separableConvolution2DNaiveEqsat.scala
umazalakain: Fixed

src/main/scala/shine/DPIA/InferAccessAnnotation.scala
umazalakain: Fixed

src/main/scala/rise/elevate/strategies/lowering.scala
umazalakain: Fixed

src/main/scala/rise/elevate/rules/traversal.scala
umazalakain: Fixed

src/main/scala/rise/core/equality.scala
Bastacyclop: I feel like having the type field straight in our primitive classes keeps preventing us from using default `equals` and `hashCode` implementations, and pushes us to use string names which is inefficient and does not feel like the rigt way. Maybe we should think about extracting the type field from our primitives. What if our expressions looked like:
```scala
sealed trait UntypedExpr
case class UntypedExprNode(node: Node[UntypedExpr])
case class TypeAnnotation(e: UntypedExpr, t: Type) extends UntypedExpr

case class TypedExpr(node: Node[TypedExpr], t: Type)

sealed trait Node[+E]
case class Var(debug: String, id: Int) extends Node[Nothing]
case class App[E](f: E, e: E) extends Node[E]
case class Lambda[E](e: E) extends Node[E]
// ...
abstract class Primitive extends Node[Nothing] {
  def typeScheme: Type
  def name: String
}
```
Then we could use the default `hashCode` and `equals` for primitives when we care only about the primitive, and not about its type, which is not currently the case.

Bastacyclop: That would also solve #152 and #134

umazalakain: Maybe I'm missing something, but in that example, if `Primitive <: Node[Nothing]`, would that not mean that in `App(f, e)` if `f : Primitive` then necessarily `e` has to be a subtype of `Node[Nothing]`?

Bastacyclop: I don't think that you will ever need a `Node[Node[_]]` so `f: Primitive` sounds unlikely to be useful. Instead you will have `f: TypedExpr` or `f: UntypedExpr` and then something like `f = TypedExpr(primitive, t)`.

Bastacyclop: You can have a look at my `typed-eqsat` branch to see expressions defined pretty much like above, and how they are used: https://github.com/rise-lang/shine/blob/a25e8b9887974d4b19dce9a4b1b6294f33feaa02/src/main/scala/rise/eqsat/Expr.scala
One of the nice aspects is that you can use that same node definition to define patterns for rewriting which can also include pattern variables on top of the regular nodes (just like we can add type annotations for untyped expressions in the example above), this would be useful if we decide to define our rewrite rule using patterns for Elevate as well.

umazalakain: Right, I got confused because I thought `Nothing` was of type `Optional[T]`, when it is in fact of type `Any`. This looks pretty good to me! Should we merge this as is and open a separate PR (it will likely take considerably more work and discussions)

Bastacyclop (approved):

PR #163, opened 2021-04-26
Created GAP8 package with Module and ModuleGenerator

bpervan:
I have created GAP8 package/module with dedicated ```Module``` and ```ModuleGenerator``` for GAP8 backend. For now ```GAP8.ModuleGenerator``` is reusing most of its functionalities from ```C.ModuleGenerator```. I have also added ```gap8``` object to ```gen.scala```. 

Please check out changes and comments in ```ModuleGenerator``` as I could use a hint or two regarding:
1. Constructing a cast statement in a more meaningful way than using an ordinary string code (line 64)
2. Constructing statements or whatever needed that would translate to something like this:
```
uint8_t* output = cl_params->output;
...
```
As you can see, currently I'm doing it in a pretty naive manner (lines 66-68)

3. Constructing a new ```ParamKind``` of type analogous to ```void*```, as assert in ```Function.scala``` fails (line 84)

Regarding ```gen.scala``` is there any chance of making ```ModuleGenerator``` a parameter of the appropriate functions (i.e. ```asStringFromExpr``` etc.), so one could specify which Module Generator they want to use, as they can for ```CodeGenerator```?

I'm aware of the drawbacks of this copy/paste model of extending code base in general :) This is a draft pull request, so feel free to suggest any other way that would improve code reuse. I was thinking of (maybe) converting C.ModuleGenerator to class which would enable me to extend it for GAP8.ModuleGenerator but that leads to inevitable refactoring of OpenCL and Cuda ```ModuleGenerator``` (which would certainly work but would break coding style used so far)

### DIFF
--- a/src/main/scala/shine/GAP8/Compilation/ModuleGenerator.scala
+++ b/src/main/scala/shine/GAP8/Compilation/ModuleGenerator.scala
@@ -1,19 +1,14 @@
 package shine.GAP8.Compilation
 
-import arithexpr.arithmetic.Cst
 import shine.C.AST.{IncludeHeader, ParamKind}
+import shine.C.Compilation.{ModuleGenerator => CModuleGenerator}
 import shine.C.{Compilation, Module}
-import shine.DPIA.Compilation.Passes._
-import shine.DPIA.Compilation.{FunDef, TranslationToImperative}
-import shine.DPIA.DSL._
+import shine.DPIA.Compilation.FunDef
 import shine.DPIA.Phrases._
 import shine.DPIA.Types._
-import shine.DPIA.primitives.functional
 import shine.{C, DPIA}
-import util.compiler.DSL.run
-import shine.C.Compilation.{ModuleGenerator => CModuleGenerator}
 
-import scala.collection.{immutable, mutable}
+import scala.collection.immutable
 
 object ModuleGenerator extends DPIA.Compilation.ModuleGenerator[FunDef] {
   override type Module = C.Module
@@ -28,7 +23,6 @@ object ModuleGenerator extends DPIA.Compilation.ModuleGenerator[FunDef] {
                   ): Phrase[ExpType] => Phrase[CommType] =
     CModuleGenerator.toImperative(gen, funDef, outParam)
 
-  //Culprit
   override def imperativeToModule(gen: CodeGenerator,
                                   funDef: FunDef,
                                   outParam: Identifier[AccType]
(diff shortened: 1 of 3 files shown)


Bastacyclop:
@bpervan For 2. you can use `shine.C.AST.VarDecl`, for example `C.AST.VarDecl("x", C.AST.Type.int, init = Some(C.AST.Literal("1")))` gives `int x = 1`.

src/main/scala/shine/GAP8/Compilation/ModuleGenerator.scala
```diff
@@ -0,0 +1,93 @@
...
+      val params = (outParam +: funDef.params).
+        map(C.AST.makeParam(C.AST.makeParamTy(gen)))
+
+      val structDecl = C.AST.StructTypeDecl(
+        "struct cluster_params",
+        params.map(i => C.AST.VarDecl(i.name, i.t))
+      )
+
+      //C.AST.Cast or sth
+      val structCastStmt = C.AST.Code("struct cluster_params* cl_params = (struct cluster_params*) args;")
```
Bastacyclop, line 64: You should be able to use `C.AST.VarDecl` with `C.AST.Cast` and a `C.AST.PointerType` of `C.AST.StructType`.

src/main/scala/shine/GAP8/Compilation/ModuleGenerator.scala
```diff
@@ -0,0 +1,93 @@
...
+      Module(
+        includes = immutable.Seq(IncludeHeader("stdint.h")),
+        decls = CModuleGenerator.collectTypeDeclarations(code, params) ++ declarations ++ Seq(structDecl),
+        functions = immutable.Seq(
+          C.AST.Function(
+            code =
+              C.AST.FunDecl(
+                funDef.name,
+                returnType = C.AST.Type.void,
+                Seq(C.AST.ParamDecl("args", C.AST.PointerType(C.AST.Type.void))),
```
Bastacyclop, line 80: I assume that you cannot use a `struct cluster_params*` pointer here for some reason?

src/main/scala/shine/GAP8/Compilation/ModuleGenerator.scala
```diff
@@ -0,0 +1,93 @@
...
+                C.AST.Block(immutable.Seq(structCastStmt, unpackingStmt, code))),
+            paramKinds =
+            //Construct new ParamKind() of type analogous to void*, assert in Function.scala fails
```
Bastacyclop, line 83: You could use `shine.DPIA.Types.OpaqueType("void*")` as a quick fix.

src/main/scala/shine/GAP8/Module.scala
```diff
@@ -0,0 +1,26 @@
...
+object Module {
+  def compose(ms: Seq[Module]): Module = ms.reduce(_ compose _)
+
+  def translateToString(m: Module): String =
+    s"""
+       |${m.includes.map(_.toString).mkString("\n")}
+       |${m.decls.map(C.AST.Printer(_)).mkString("\n")}
+       |${m.functions.map(f => C.AST.Printer(f.code)).mkString("\n")}
+       |""".stripMargin
+}
```
Bastacyclop, line 26: this file is dead code, right?

src/main/scala/shine/GAP8/Compilation/ModuleGenerator.scala
bpervan: If you are refering to ```void* args``` this is actually fine as that's how low-level interface looks like, i.e.:
```C
void cluster_core_task(void* args)
```

src/main/scala/shine/GAP8/Module.scala
bpervan: Yes it is, I see now, I'm gonna refactor this slightly

Bastacyclop:
I wonder if instead of having your own module generator you could instead design this as some kind of module transformer, say:
```scala
package shine.gap8

object generateInterface {
  def apply(functionName: String, wrappedName: String, m: C.Module): C.Module = {
    val function = m.functions.find(_.name == functionName)
    val wrapped = C.AST.Function(/* ... from void* args to calling the function ... */)
    m.copy(functions = m.functions +: wrapped)
  }
}
```

bpervan:
> I wonder if instead of having your own module generator you could instead design this as some kind of module transformer

To be honest, maybe it is better, but I don't know yet. I see your point though. The idea was to create another type of Module, as module for GAP8 will at some point in future start looking slightly different than C module (at least I think it will), having host code, other various declarations etc. This is something we should discuss during our meeting @michel-steuwer

PR #166, opened 2021-05-03
GAP8 module

bpervan:
I ditched the original idea (#163) of having a dedicated ```GAP8``` ```ModuleGenerator``` and instead implemented a converter from an already generated  ```C.Module``` to ```GAP8.Module```. ```GAP8.Module``` has also been refactored as it will eventually hold host code and accelerator functions (at some point in future). The ```gen.scala``` has been updated accordingly, to utilize the converter.


src/main/scala/shine/GAP8/Module.scala
```diff
@@ -0,0 +1,94 @@
...
+            Seq(structCastDecl)
+              ++ unpackingStmts
+              ++ accFunction.map(f => Seq(f)).getOrElse(Seq()).map(_.code.body)
+          )
+        ),
+      paramKinds =
+        Seq(ParamKind(DPIA.Types.OpaqueType("void*"), C.AST.ParamKind.Kind.input))
+    )
+
+    Module(
```
michel-steuwer, line 85: Can we add a TODO here, highlighting that the generated module is an accelerator module and that the host code is still to come?

michel-steuwer (changes requested):

michel-steuwer (approved):
👍

PR #174, opened 2021-05-13
Do not rely on any explicitness information for inference and constraint resolution

umazalakain:
Currently we use a global set of type variables to `preserve`, but we still rely on explicitness information to preserve bound variables. Given that we have globally unique variable names we can simply add every bound variable to the set of variables to preserve. The `preserve` set needs to be modified in recursive calls to `solveOne` (we currently don't have to do that because we use `_.asImplicit` on some variables instead).

Fragment kinds and matrix layouts are now traversed too and their identifiers are correctly picked up. These identifiers are however ignored by `makeClosed`, because they cannot be bound.

`IsClosedForm` now correctly traverses lambdas, it used to skip their types.

It doesn't seem to make a noticeable impact on the performance of the test suite.

### DIFF
--- a/src/main/scala/rise/core/DSL/infer.scala
+++ b/src/main/scala/rise/core/DSL/infer.scala
@@ -43,12 +43,8 @@ object infer {
     val (e_preserve, e_wo_assertions) = traverse(e, collectPreserve)
     // Collect constraints
     val (typed_e, constraints) = constrainTypes(exprEnv)(e_wo_assertions)
-    // Collect free variables both before and after constraint gathering:
-    // Some types in the collected constraints might not be present after constraint gathering (Fixme: BUG?)
-    val ftvs = IsClosedForm.freeVars(e_wo_assertions)._2.toSet ++ IsClosedForm.freeVars(typed_e)._2.toSet
-    val toSubstitute = (ftvs removedAll e_preserve) removedAll typeEnv
     // Solve constraints while preserving the FTVs in preserve
-    val solution = Constraint.solve(constraints, toSubstitute, Seq())
+    val solution = Constraint.solve(constraints, e_preserve ++ typeEnv, Seq())
     // Apply the solution
     val res = traverse(typed_e, Visitor(solution))
     if (printFlag == Flags.PrintTypesAndTypeHoles.On) {
--- a/src/main/scala/rise/core/DSL/infer.scala
+++ b/src/main/scala/rise/core/DSL/infer.scala
@@ -99,14 +99,32 @@ object infer {
 
   private val collectPreserve = new PureAccumulatorTraversal[Set[Kind.Identifier]] {
     override val accumulator = SetMonoid
+
+    override def typeIdentifier[I <: Kind.Identifier]: VarType => I => Pair[I] = _ => {
+      case i: Kind.Explicitness =>
+        accumulate(if (i.isExplicit) Set(i) else Set())(i.asInstanceOf[I])
+      case i => accumulate(Set())(i)
+    }
+
+    override def nat: Nat => Pair[Nat] = ae => {
+      val pres = mutable.ListBuffer[Kind.Identifier]()
+      val r = ae.visitAndRebuild({
+        case i: NatIdentifier if i.isExplicit => pres += i; i
+        case n => n
+      })
+      accumulate(pres.toSet)(r)
+    }
+
     override def expr: Expr => Pair[Expr] = {
       // Transform assertions into annotations, collect FTVs
       case TypeAssertion(e, t) =>
-        val (s, e1) = expr(e).unwrap
-        accumulate(s ++ getFTVs(t))(TypeAnnotation(e1, t) : Expr)
+        val (s1, e1) = expr(e).unwrap
+        val (s2, _) = `type`(t).unwrap
+        accumulate(s1 ++ s2 ++ getFTVs(t))(TypeAnnotation(e1, t) : Expr)
       // Collect FTVs
       case Opaque(e, t) =>
-        accumulate(getFTVs(t).toSet)(Opaque(e, t) : Expr)
+        val (s, _) = `type`(t).unwrap
+        accumulate(s ++ getFTVs(t).toSet)(Opaque(e, t) : Expr)
       case e => super.expr(e)
     }
   }
--- a/src/main/scala/rise/core/DSL/infer.scala
+++ b/src/main/scala/rise/core/DSL/infer.scala
@@ -35,9 +35,8 @@ object infer {
   }
 
   type ExprEnv = Map[String, Type]
-  type TypeEnv = Set[Kind.Identifier]
 
-  def apply(e: Expr, exprEnv: ExprEnv = Map(), typeEnv : TypeEnv = Set(),
+  def apply(e: Expr, exprEnv: ExprEnv = Map(), typeEnv : Set[Kind.Identifier] = Set(),
             printFlag: Flags.PrintTypesAndTypeHoles = Flags.PrintTypesAndTypeHoles.Off,
             explDep: Flags.ExplicitDependence = Flags.ExplicitDependence.Off): Expr = {
     // TODO: Get rid of TypeAssertion and deprecate, instead evaluate !: in place and use `preserving` directly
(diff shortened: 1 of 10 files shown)


src/main/scala/rise/core/types/Constraints.scala
```diff
@@ -132,19 +131,18 @@ object Constraint {
...
                 decomposed(
                   Seq(
-                    NatConstraint(n, na.asImplicit),
-                    NatConstraint(n, nb.asImplicit),
+                    NatConstraint(n, na),
+                    NatConstraint(n, nb),
                     TypeConstraint(ta, tb)
-                  )
-                )
+                  ), preserve - na - nb)
```
Bastacyclop, line 145: `+ n`? same in other spots. I wonder if we could also alternatively directly apply the required substitutions (`na -> n` and `nb -> n`) where necessary and only generate 1 constraint instead of 3 here. Something like:
```scala
val n = NatIdentifier(freshName("n"))
decomposed(Seq(TypeConstraint(
  substitute.natsInType(ta, Map(na -> n)),
  substitute.natsInType(tb, Map(nb -> n))
))), preserve + n)
```

Bastacyclop (commented):
Looks like a good solution to me overall. I think that we should unify the terminology around "preserve"/"bound"/"scope"/"environment" in the code and variable names. For the `explicitlyDependent` stuff maybe you should talk to @fedepiz .

PS: what tests are currently failing?

src/main/scala/rise/core/types/Constraints.scala
umazalakain: Yup, I was just picking up on that, I will get back to you!

umazalakain:
@Bastacyclop Regarding the naming, I think it would make sense to use "typing environment" or "typing context" on the infer side: it is more than a scope, since it has kinding information; it is in line with using "exprEnv" for the expression world, and we are not yet concerned about how this typing environment is used. On the constraints side I would switch to "preserve", since we add and remove things to it regardless of any typing contexts.

I added the explicit variables that we create to the preserve set (I think it's best if we leave making any optimizations for later on once the tests pass). There are 13 tests that don't pass, I will send you a file with the results on zulip.

Re the dependent types @fedepiz, I tried collecting the identifiers that are made implicit into a set and then subtracting those from the preserve set in lines 133-135, but it didn't seem to help.

fedepiz:
One quick remark about explicit dependence. I have in recent times started to phase it out in my work. and prefer instead throwing the compiler more handwritten constraints until it's happy. That's because it can only infer some marginally more complex situations, but in the end, writing out one type of signature can save your bacon much more reliably.

So, if the `explicit dependence` stuff is getting in the way, please feel free to throw the mechanism and the few associated tests that might fail if you don't feel like providing type signatures for some alien-looking tests. As long as most of the basic dependent stuff is in place, then I am more than happy with it.

umazalakain:
Thomas and I had a meeting and identified some of the possible things that might have been going wrong (there are about 10 test cases that don't pass currently) and talked about a possible roadmap:

- Solutions should not substitute within opaques. Moreover, solutions should not contain type variables of opaques as keys. We should add an assertion that makes sure of this.
- At some point we can remove the explicitly dependent stuff if it gets in the way.
- Rewrites need to happen a la eqsat: get the expression and type environment from the LHS, give those to the inference that is being ran on the RHS. This removes the need for type assertions.
- There is no need for per-subconstraint environment information in `solveOne`, preservation information should be shared among subcontraints.
- What happens to the environment information attached to a constraint when substitutions are applied to that constraint? Substitutions should also be applied to the environment information.
- Does the application of solutions need to be scoped? Add the following test case:
  ```
  ((dt1 : data) -> _x) _x
  ((dt2 : data) -> dt2) _x
  ```
  inference creates a new type variable `dt` and substitutes `dt1 |-> dt` and `dt2 |-> dt`:
  ```
  ((dt : data) -> _x) _x
  ((dt : data) -> dt) _x
  ```
  then inference creates a constraint `_x ~ dt` resulting in the substitution `_x |-> dt`. applying this substitution globally gives
  ```
  ((dt : data) -> dt) dt
  ((dt : data) -> dt) dt
  ```
  however note that here the function argument `dt` is not bound: the expression is ill-scoped.

src/main/scala/rise/core/IsClosedForm.scala
```diff
@@ -6,15 +6,32 @@ import rise.core.traverse._
 import rise.core.types._
 
 object IsClosedForm {
+  case class OrderedSet[T](ordered : Seq[T], unique : Set[T])
```
Bastacyclop, line 9: Is `unique` a good name? Writing the predicate `unique(x)` reads like "is x unique?", which seems wrong. What about simply `set`?

src/main/scala/rise/core/IsClosedForm.scala
```diff
@@ -6,15 +6,32 @@ import rise.core.traverse._
...
+  object OrderedSet {
+    def one[T] : T => OrderedSet[T] = t => OrderedSet(Seq(t), Set(t))
+    def add[T] : T => OrderedSet[T] => OrderedSet[T] = t => ts =>
+      if (ts.unique.contains(t)) ts else OrderedSet(t +: ts.ordered, ts.unique + t)
+    def empty[T] : OrderedSet[T] = OrderedSet(Seq(), Set())
+    def append[T] : OrderedSet[T] => OrderedSet[T] => OrderedSet[T] = x => y => {
+      val ordered = x.ordered.filter(!y.unique.contains(_)) ++ y.ordered
```
Bastacyclop, line 16: When writing `x append y`, should `x` not have priority on ordering? So that `[1 2 4] append [3 2] = [1 2 4 3]`? This would correspond to `x.ordered ++ y.ordered.filter(!x.unique(_)))`

src/main/scala/rise/core/DSL/infer.scala
```diff
@@ -83,74 +71,55 @@ object infer {
...
+      case _ => return_
+    }
+
     override def expr: Expr => Pair[Expr] = {
       // Transform assertions into annotations, collect FTVs
       case TypeAssertion(e, t) =>
-        val (s, e1) = expr(e).unwrap
-        accumulate(s ++ getFTVs(t))(TypeAnnotation(e1, t) : Expr)
+        val (s1, e1) = expr(e).unwrap
+        accumulate(s1 ++ IsClosedForm.varsToClose(t))(TypeAnnotation(e1, t) : Expr)
```
Bastacyclop, line 86: What is the issue with using `freeVars`?

src/main/scala/rise/eqsat/NamedRewrite.scala
```diff
@@ -9,18 +9,18 @@ object NamedRewrite {
...
       assert(t == rct.TypePlaceholder)
       name -> rct.TypeIdentifier("t" + name)
     }
-    val typedLhs = preservingWithEnv(lhs, untypedFreeV, Set())
-    val freeV = collectFreeEnv(typedLhs)
-    val (_, freeT) = rise.core.IsClosedForm.freeVars(typedLhs)
-    val typedRhs = preservingWithEnv(rc.TypeAnnotation(rhs, typedLhs.t), freeV, freeT)
+    val typedLhs = infer(lhs, untypedFreeV, Set())
+    val freeV = infer.collectFreeEnv(typedLhs)
+    val freeT = rise.core.IsClosedForm.freeVars(typedLhs)._2.toSet
```
Bastacyclop, line 22: We could potentially get the `Set` from the `OrderedSet` to avoid extra work here.

Bastacyclop (commented):
Looks good to me. Maybe we can open relevant issues for the points discussed in https://github.com/rise-lang/shine/pull/174#issuecomment-841278331 ?

src/main/scala/rise/core/IsClosedForm.scala
umazalakain: thanks, done! for consistency I renamed `unique` to `set` and `ordered` to `seq`

src/main/scala/rise/core/DSL/infer.scala
umazalakain: Indeed we should be calling `freeVars` here, we don't want to skip fragment kind and matrix layout identifiers, thanks

umazalakain:
@Bastacyclop will do, thanks for the review!

umazalakain:
> Looks good to me. Maybe we can open relevant issues for the points discussed in [#174 (comment)](https://github.com/rise-lang/shine/pull/174#issuecomment-841278331) ?

I created issues for the first #177 and last #178 points, the rest has been resolved

Bastacyclop (approved):

PR #176, opened 2021-05-28
Removes all explicitness information in RISE

umazalakain:
Builds on top of #174

### DIFF
--- a/src/main/scala/rise/core/makeClosed.scala
+++ b/src/main/scala/rise/core/makeClosed.scala
@@ -9,7 +9,7 @@ object makeClosed {
     val (expr, ts) = IsClosedForm.varsToClose(e)._2.foldLeft((e, Map[Type, Type]()))((acc, ftv) => acc match {
       case (expr, ts) => ftv match {
         case i: TypeIdentifier =>
-          val dt = DataTypeIdentifier(freshName("dt"), isExplicit = true)
+          val dt = DataTypeIdentifier(freshName("dt"))
           (DepLambda[DataKind](dt, expr)(DepFunType[DataKind, Type](dt, expr.t)), (ts ++ Map(i -> dt)))
         case i: DataTypeIdentifier =>
           (DepLambda[DataKind](i, expr)(DepFunType[DataKind, Type](i, expr.t)), ts)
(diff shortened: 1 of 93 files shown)


src/main/scala/rise/core/makeClosed.scala
```diff
@@ -6,36 +6,24 @@ import rise.core.types._
...
-      case (expr, (ts, ns, as, n2ds)) => ftv match {
+    val (_, ftvs) = IsClosedForm.varsToClose(e)
+    val (expr, ts) = ftvs.foldLeft((e, Map[Type, Type]()))((acc, ftv) => acc match {
+      case (expr, ts) => ftv match {
         case i: TypeIdentifier =>
-          val dt = DataTypeIdentifier(freshName("dt"), isExplicit = true)
-          (DepLambda[DataKind](dt, expr)(DepFunType[DataKind, Type](dt, expr.t)),
-            (ts ++ Map(i -> dt), ns, as , n2ds))
+          val dt = DataTypeIdentifier(freshName("dt"))
+          (DepLambda[DataKind](dt, expr)(DepFunType[DataKind, Type](dt, expr.t)), (ts ++ Map(i -> dt)))
```
Bastacyclop, line 14: Do we have tests that rely on this code path? We might want to allow dependent lambdas binding types in the future, changing to a data type identifiers seems like a hack we should get rid of.

Bastacyclop (commented):
Nice :+1:

src/main/scala/rise/core/makeClosed.scala
umazalakain: That seems to be the case. Initially things blew up because some traversal paths for dependent lambdas on type identifiers were missing, once I added those the trouble seemed to be with the fission tests in the fissionFussion suite and the simple fusion test in the traversals suite: they use `mapLastFission`, which only works on data types, weakening this condition breaks inference on the rewritten term. I will open a ticket.

michel-steuwer (approved):
👍

michel-steuwer:
I merged master into this branch already: https://github.com/rise-lang/shine/commit/a95a9a7e465ae1acb171209acbd30b0435cf9f05

Let's wait a bit to see if/when this trickles down here to the PR ...

michel-steuwer:
I have opened a ticket with GitHub support for them to investigate why the commits to this branch do not show up here.

umazalakain:
@michel-steuwer yes, something weird is going on. For now I tried to push to master directly, but it's a protected branch

PR #181, opened 2021-06-01
Removed type projection from Kinds

michel-steuwer:
This PR prepares the migration to Scala 3 by removing the generic type projection from `Kind`s.

`DepLambda`, `DepApp`, `DepFunType`, and `DepPairType` now all carry the kind as a plain data member.
This also allows an easier uniform treatment in many pattern matching cases.


Bastacyclop:
I wonder if we could not simplify this? Maybe something with much less generics like https://scastie.scala-lang.org/VS9nRpmpReWli1IY9M7yBA

michel-steuwer:
> I wonder if we could not simplify this? Maybe something with much less generics like [scastie.scala-lang.org/VS9nRpmpReWli1IY9M7yBA](https://scastie.scala-lang.org/VS9nRpmpReWli1IY9M7yBA)

The main problem that I see with your solution is that now all Kinds must inherit from a single trait (`Kind.Value`). That means that `ArithExpr` etc. must inherit from that.

Bastacyclop (approved):
We can also revisit this later on.

PR #182, opened 2021-06-03
Add documentation for type system

michel-steuwer:
This PR adds documentation for the type system of RISE. It looks like:

image

The `latex.scala` file in the `rise.core.types` package constructs RISE types that are then printed in LaTeX.

Once we have reconciled the Kinds across RISE and DPIA, we should add the `Fragment` and `MatrixLayout` kinds to the documentation (as indicated by the commented out code).

### DIFF
--- a/docs-website/docusaurus.config.js
+++ b/docs-website/docusaurus.config.js
@@ -1,13 +1,25 @@
+const math = require('remark-math');
+const katex = require('rehype-katex');
 module.exports = {
   title: 'RISE Language Documentation',
   tagline: 'A functional pattern-based data-parallel language',
   url: 'https://rise-lang.org',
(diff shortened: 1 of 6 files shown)


umazalakain (approved):
This is amazing <3

Bastacyclop:
@michel-steuwer Is the plan for this to replace https://github.com/rise-lang/doc ?

michel-steuwer:
> @michel-steuwer Is the plan for this to replace [rise-lang/doc](https://github.com/rise-lang/doc) ?

Yes!
[rise-lang/doc](https://github.com/rise-lang/doc) is always out of date, so I hope that by integrating the documentation in our code, it will be easier to keep it up to date.

PR #183, opened 2021-06-05
Rework rise's type identifiers; alias rise's `NatIdentifier`s to arithexpr's `NamedVar`

umazalakain:
This PR gets eliminates the need to subclass `Kind.Identifier` when creating something that is meant to be considered a type identifier. Instead, it uses `Kind.Identifier` as a closed trait where each constructor accepts as its only argument an identifier of the desired type. It is then the kinding system which relates the two with `Kind[+T, +I, KI <: Kind.Identifier]`. Two functions `toIdentifier` and `fromIdentifier` are provided that box and unbox the identifiers into and out of `Kind.Identifier`. These two functions and `idName` are defined as functions instead of as methods in the `Kind` trait so that we can have a covariant `+I`.

Whenever a set of identifiers of different kinds is required, `Kind.Identifier` can be used. In its current form, there is some boxing-unboxing overhead in traversals, but I think this can be avoided.

As a result of this PR, variables in arithexpr and variables in rise are one and the same: `NamedVar`. The idea is to do something similar for dpia, thus having a uniform `NamedVar` all across the codebase.

### DIFF
--- a/src/main/scala/rise/core/traverse.scala
+++ b/src/main/scala/rise/core/traverse.scala
@@ -3,6 +3,7 @@ package rise.core
 import scala.language.implicitConversions
 import util.monads._
 import rise.core.semantics._
+import rise.core.types.Kind.{IAddressSpace, IDataType, IFragmentKind, IMatrixLayout, INat, INatToData, INatToNat, IType}
 import rise.core.types._
 
 object traverse {
@@ -23,29 +24,29 @@ object traverse {
       for { t1 <- `type`(i.t)}
         yield i.setType(t1).asInstanceOf[I]
     def typeIdentifierDispatch[I <: Kind.Identifier] : VarType => I => M[I] = vt => i => (i match {
-      case n: NatIdentifier => bind(typeIdentifier(vt)(n))(nat)
-      case dt: DataTypeIdentifier => bind(typeIdentifier(vt)(dt))(datatype)
-      case a: AddressSpaceIdentifier => bind(typeIdentifier(vt)(a))(addressSpace)
-      case m: MatrixLayoutIdentifier => bind(typeIdentifier(vt)(m))(matrixLayout)
-      case f: FragmentKindIdentifier => bind(typeIdentifier(vt)(f))(fragmentKind)
-      case n2n: NatToNatIdentifier => bind(typeIdentifier(vt)(n2n))(natToNat)
-      case n2d: NatToDataIdentifier => bind(typeIdentifier(vt)(n2d))(natToData)
-      case t: TypeIdentifier => typeIdentifier(vt)(t)
+      case t: IType => typeIdentifier(vt)(t)
+      case n: INat => bind(bind(typeIdentifier(vt)(n))(i => nat(i.id)))(i => return_(INat(i.asInstanceOf[NatIdentifier])))
+      case dt: IDataType => bind(bind(typeIdentifier(vt)(dt))(i => datatype(i.id)))( i => return_(IDataType(i.asInstanceOf[DataTypeIdentifier])))
+      case a: IAddressSpace => bind(bind(typeIdentifier(vt)(a))(i => addressSpace(i.id)))( i => return_(IAddressSpace(i.asInstanceOf[AddressSpaceIdentifier])))
+      case m: IMatrixLayout => bind(bind(typeIdentifier(vt)(m))(i => matrixLayout(i.id)))( i => return_(IMatrixLayout(i.asInstanceOf[MatrixLayoutIdentifier])))
+      case f: IFragmentKind => bind(bind(typeIdentifier(vt)(f))(i => fragmentKind(i.id)))( i => return_(IFragmentKind(i.asInstanceOf[FragmentKindIdentifier])))
+      case n2n: INatToNat => bind(bind(typeIdentifier(vt)(n2n))(i => natToNat(i.id)))( i => return_(INatToNat(i.asInstanceOf[NatToNatIdentifier])))
+      case n2d: INatToData => bind(bind(typeIdentifier(vt)(n2d))(i => natToData(i.id)))( i => return_(INatToData(i.asInstanceOf[NatToDataIdentifier])))
     }).asInstanceOf[M[I]]
     def natDispatch : VarType => Nat => M[Nat] = vt => {
-      case i : NatIdentifier => bind(typeIdentifier(vt)(i))(nat)
+      case i : NatIdentifier => bind(typeIdentifier(vt)(INat(i)))(i => nat(i.id))
       case n => nat(n)
     }
     def matrixLayoutDispatch : VarType => MatrixLayout => M[MatrixLayout] = vt => {
-      case i : MatrixLayoutIdentifier => bind(typeIdentifier(vt)(i))(matrixLayout)
+      case i : MatrixLayoutIdentifier => bind(typeIdentifier(vt)(IMatrixLayout(i)))(i => matrixLayout(i.id))
       case m => matrixLayout(m)
     }
     def fragmentKindDispatch : VarType => FragmentKind => M[FragmentKind] = vt => {
-      case i : FragmentKindIdentifier => bind(typeIdentifier(vt)(i))(fragmentKind)
+      case i : FragmentKindIdentifier => bind(typeIdentifier(vt)(IFragmentKind(i)))(i => fragmentKind(i.id))
       case m => fragmentKind(m)
     }
     def dataTypeDispatch : VarType => DataType => M[DataType] = vt => {
-      case i : DataTypeIdentifier => bind(typeIdentifier(vt)(i))(datatype)
+      case i : DataTypeIdentifier => bind(typeIdentifier(vt)(IDataType(i)))(i => datatype(i.id))
       case d => datatype(d)
     }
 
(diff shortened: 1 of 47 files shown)


src/main/scala/rise/core/traverse.scala
```diff
@@ -23,29 +24,29 @@ object traverse {
...
-      case t: TypeIdentifier => typeIdentifier(vt)(t)
+      case t: IType => typeIdentifier(vt)(t)
+      case n: INat => bind(bind(typeIdentifier(vt)(n))(i => nat(i.id)))(i => return_(INat(i.asInstanceOf[NatIdentifier])))
+      case dt: IDataType => bind(bind(typeIdentifier(vt)(dt))(i => datatype(i.id)))( i => return_(IDataType(i.asInstanceOf[DataTypeIdentifier])))
+      case a: IAddressSpace => bind(bind(typeIdentifier(vt)(a))(i => addressSpace(i.id)))( i => return_(IAddressSpace(i.asInstanceOf[AddressSpaceIdentifier])))
+      case m: IMatrixLayout => bind(bind(typeIdentifier(vt)(m))(i => matrixLayout(i.id)))( i => return_(IMatrixLayout(i.asInstanceOf[MatrixLayoutIdentifier])))
+      case f: IFragmentKind => bind(bind(typeIdentifier(vt)(f))(i => fragmentKind(i.id)))( i => return_(IFragmentKind(i.asInstanceOf[FragmentKindIdentifier])))
+      case n2n: INatToNat => bind(bind(typeIdentifier(vt)(n2n))(i => natToNat(i.id)))( i => return_(INatToNat(i.asInstanceOf[NatToNatIdentifier])))
+      case n2d: INatToData => bind(bind(typeIdentifier(vt)(n2d))(i => natToData(i.id)))( i => return_(INatToData(i.asInstanceOf[NatToDataIdentifier])))
```
Bastacyclop, line 34: Why do we need casts here now?

src/main/scala/rise/core/types/Kinds.scala
```diff
@@ -1,39 +1,77 @@
...
-sealed trait Kind[+T, +I <: Kind.Identifier] {
+sealed trait Kind[+T, +I, KI <: Kind.Identifier] {
   def name :String
 }
 
+
 object Kind {
-  trait Identifier {
-    def name: String
+  trait Identifier { def name: String }
```
michel-steuwer, line 9: In Scala 3 maybe this could be written using Match Types: https://dotty.epfl.ch/docs/reference/new-types/match-types.html

src/main/scala/rise/core/types/Kinds.scala
```diff
@@ -1,39 +1,77 @@
...
+  def fromIdentifier[T, I, KI <: Kind.Identifier](kind : Kind[T, I, KI], i : KI) : I = kind match {
+    case TypeKind => i.asInstanceOf[IType].id.asInstanceOf[I]
+    case DataKind => i.asInstanceOf[IDataType].id.asInstanceOf[I]
+    case NatKind => i.asInstanceOf[INat].id.asInstanceOf[I]
+    case AddressSpaceKind => i.asInstanceOf[IAddressSpace].id.asInstanceOf[I]
+    case NatToNatKind => i.asInstanceOf[INatToNat].id.asInstanceOf[I]
+    case NatToDataKind => i.asInstanceOf[INatToData].id.asInstanceOf[I]
+    case NatCollectionKind => i.asInstanceOf[INatCollection].id.asInstanceOf[I]
   }
 }
```
Bastacyclop, line 49: Would it not save on some casting to match on the identifier instead of the kind object?

src/main/scala/rise/core/types/Solution.scala
```diff
@@ -141,18 +139,18 @@ case class Solution(ts: Map[Type, Type],
     case FragmentTypeConstraint(a, b) => FragmentTypeConstraint(apply(a), apply(b))
     case NatToDataConstraint(a, b) => NatToDataConstraint(apply(a), apply(b))
     case NatCollectionConstraint(a, b) => NatCollectionConstraint(apply(a), apply(b))
-    case DepConstraint(NatKind, df, arg: Nat, t) => DepConstraint(NatKind, apply(df), apply(arg), apply(t))
-    case DepConstraint(DataKind, df, arg: DataType, t) =>
+    case DepConstraint(_, df, arg: Nat, t) => DepConstraint(NatKind, apply(df), apply(arg), apply(t))
```
michel-steuwer, line 142: Is there a reason why you removed this?

src/main/scala/rise/core/types/Solution.scala
```diff
@@ -91,7 +91,6 @@ case class Solution(ts: Map[Type, Type],
       case lambda@NatToDataLambda(x, body) =>
         val xSub = apply(x) match {
           case n: NatIdentifier => n
-          case n: arithexpr.arithmetic.NamedVar => NatIdentifier(n.name, n.range)
```
michel-steuwer, line 94: Nice!

Bastacyclop (commented):
Looks good overall, although like for #181, I wish we could have less generic parameters. It feels like the kind should imply everything else (`T`, `I`, `KI`), not depend on it.

src/main/scala/rise/elevate/rules/algorithmic.scala
```diff
@@ -58,7 +58,7 @@ object algorithmic {
 
   // padEmpty n >> padEmpty m -> padEmpty n + m
   @rule def padEmptyFusion: Strategy[Rise] = {
-    case e @ App(DepApp(NatKind, padEmpty(), m: Nat), App(DepApp(NatKind, padEmpty(), n: Nat), in)) =>
+    case e @ App(DepApp(_, padEmpty(), m: Nat), App(DepApp(_, padEmpty(), n: Nat), in)) =>
```
michel-steuwer, line 61: Same question here: why did you drop the explicit kinds in the matching everywhere?

src/main/scala/rise/elevate/rules/package.scala
```diff
@@ -16,7 +16,7 @@ package object rules {
     case App(Lambda(x, b), v) =>
       Success(substitute.exprInExpr(v, `for` = x, in = b))
     case DepApp(k1, DepLambda(k2, x, b), v) if k1 == k2 =>
-      Success(substitute.kindInExpr(k1, v, `for` = x, in = b))
+      Success(substitute.kindInExpr(k2, v, `for` = x, in = b))
```
michel-steuwer, line 19: this shouldn't change anything (as `k1 == k2`). why the change?

michel-steuwer (changes requested):
Looks overall good to me.

@umazalakain could you please comment on the individual points I highlighted.

src/main/scala/rise/core/traverse.scala
umazalakain: Before it was sufficient with the outer cast `.asInstanceOf[M[I]]`, now because we need to wrap it with the pertinent `Kind.Identifier` constructor we need to cast once in there, once outside (I should check if the outside one can be removed now though)

src/main/scala/rise/core/types/Kinds.scala
umazalakain: We could do that, but then the match wouldn't be closed (since `I` is not a closed type). If Scala's type system was a little bit more powerful, it could infer that the match on `kind` uniquely determines `I`, and thus the cast wouldn't be necessary. This might be a thing in Scala 3?

src/main/scala/rise/core/types/Kinds.scala
umazalakain: Probably, yes!

src/main/scala/rise/core/types/Solution.scala
umazalakain: No, thanks for catching this, must have gone by in the many edits

src/main/scala/rise/elevate/rules/algorithmic.scala
umazalakain: Ah, I should roll back this too. On my first approach the `I` parameter to `Kind` was not covariant (since `idName`, and `toIdentifier` were methods in `Kind` instead of functions outside of it), and Scala complained about matching on the `kind`. I believe this should work now though, I will try to roll back.

src/main/scala/rise/elevate/rules/package.scala
umazalakain: This is because Scala's type system has not internalised the fact that `k1 == k2`, and thus `x` is of the type `I` bound in `k2`, not the one bound in `k1`.

src/main/scala/rise/core/traverse.scala
umazalakain: > Looks good overall, although like for #181, I wish we could have less generic parameters. It feels like the kind should imply everything else (`T`, `I`, `KI`), not depend on it.

I totally agree, these type parameters leak out everywhere, it's pretty annoying (especially because they leak into some of the `Expr` constructors).

umazalakain:
> Looks good overall, although like for #181, I wish we could have less generic parameters. It feels like the kind should imply everything else (`T`, `I`, `KI`), not depend on it.

I totally agree, these type parameters leak out everywhere, it's pretty annoying (especially because they leak into some of the Expr constructors).

src/main/scala/rise/core/traverse.scala
umazalakain: Nope, Scala can't see that matching `I` against e.g. `INatToData` and determines `I` to be `INatToData` on the rhs of the case expression, and that therefore `return_(INatToData(_))` is ok. However what we can do is route through `nat`/`datatype`/etc first, then through `typeIdentifier`, that saves us from an extra bind, and the tests seem to pass :)

src/main/scala/rise/core/types/Solution.scala
umazalakain: Restored

src/main/scala/rise/elevate/rules/algorithmic.scala
umazalakain: Restored

umazalakain:
@michel-steuwer @Bastacyclop This should be ready to merge now

michel-steuwer (approved):
👍

PR #186, opened 2021-06-11
Runtime interface reworked, take #2

bpervan:
Here is a second take on the runtime interface

In short:
1. Functions with platform-specific parameters are moved to platform-specific header files
2. ```runtime.h``` is rearranged a bit, as some declarations have to come before inclusion of the platform-specific header files which use those declarations
3. ```KARG``` and ```LARG``` macros are moved to ```runtime.h```
3. Implementation files are separated to respective subfolders
4. Paths in [ExecuteOpenCL.scala](src/main/scala/util/ExecuteOpenCL.scala) are updated to conform the new folder structure
5. Defines are added to appropriate Modules which specify target platform, just as a proposal


runtime/gap8/gap8.h
```diff
@@ -0,0 +1,31 @@
...
+typedef void* DeviceBuffer;
+
+Context createContext(int device_id);
+Kernel loadKernel(void (*handler)(void*), size_t stack_size);
+void launchKernel(
+  Context ctx, Kernel k,
+  int num_threads,
+  size_t arg_count, const KernelArg* args);
+
+#endif
```
Bastacyclop, line 31: Maybe this file could be named `gap8/runtime.h`, and we would have the generic `runtime.h` and then `ocl/runtime.h`, etc.

runtime/runtime.h
```diff
@@ -20,29 +32,17 @@ typedef enum {
...
-  size_t size;
-  void* value;
-} KernelArg;
-
-void launchKernel(
-  Context ctx, Kernel k,
-  const size_t* global_size, const size_t* local_size,
-  size_t arg_count, const KernelArg* args);
+#define KARG(val) { sizeof(val), &val }
+#define LARG(size) { size, NULL }
```
Bastacyclop, line 46: I do not understand why these macros are moved here, are they meaningful for gap8?

runtime/runtime.h
```diff
@@ -5,13 +5,25 @@
...
-#include "time_utils.h"
 
 typedef struct ContextImpl* Context;
 typedef struct KernelImpl* Kernel;
 typedef struct BufferImpl* Buffer;
 
+typedef struct {
+  size_t size;
+  void* value;
+} KernelArg;
```
Bastacyclop, line 16: Why is this in `runtime.h`? It is useful for `launchKernel` which is defined by the backend. I would move this to the specific backends.

runtime/runtime.h
```diff
@@ -5,13 +5,25 @@
...
+
+#ifndef SHINE_TARGET_GAP8
+    #include "ocl/ocl.h"
+#else
+    #include "gap8/gap8.h"
+#endif
+#include "time_utils.h"
```
Bastacyclop, line 23: I would rather have the backend-specific headers include the generic header, and the user directly include the backend-specific header (instead of including the generic one and having to define a macro to instantiate it).

Bastacyclop (commented):
Here are some fresh comments.

src/main/scala/util/ExecuteOpenCL.scala
```diff
@@ -8,9 +8,10 @@ object ExecuteOpenCL {
   case class Exception(msg: String) extends Throwable
 
   val runtimePath = "runtime/"
+  val platformPath = "runtime/ocl/"
   val executorHeadersPath = "lib/executor/lib/Executor/include/"
   val libs = "-lm -lOpenCL"
-  val includes = s"-I$runtimePath -I$executorHeadersPath"
+  val includes = s"-I$runtimePath -I$platformPath -I$executorHeadersPath"
```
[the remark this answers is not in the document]
bpervan: Include does not recurse by default, we need to edit ```runtimePath``` then if we decide to keep this new file hierarchy in ```runtime``` folder (probably just by concatenating ```/**``` though)

runtime/gap8/gap8.h
bpervan: I would personally prefer to additionally differentiate those files by naming them differently, just to avoid any potential confusion, but it's a matter of preference I guess

runtime/runtime.h
bpervan: I reused ```KARG``` macro, probably moved ```LARG``` with it to mutual interface by inertia. Anyways, ```KARG``` is meaningful for GAP8,

runtime/runtime.h
bpervan: Right now I reused ```struct KernelArg``` to pass arguments to a function which is 'kernel' in my case

runtime/runtime.h
bpervan: Right, I just followed the current paradigm, but backend-specific headers including generic header sound better for me as well :+1:

src/main/scala/util/ExecuteOpenCL.scala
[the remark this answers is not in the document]
Bastacyclop: I do not think that it is necessary/desirable for it to recurse (i.e. no need for `runtime/ocl/` to be in the include paths). You can still use paths relative to the runtime root as in `#include "gap8/gap8.h"` or  `#include "runtime.h` instead of ` #include "gap8.h"` or `#include "../runtime.h"`.

runtime/gap8/gap8.h
Bastacyclop: We could also consider to pull it out of the folder as in `runtime/gap8.h`?

src/main/scala/util/ExecuteOpenCL.scala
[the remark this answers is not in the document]
bpervan: True, I misunderstood this, thinking that you referred to implementation files as well

runtime/gap8/gap8.h
bpervan: I've left it like this for now. We can rework this later if needed

runtime/runtime.h
bpervan: Moved ```struct KernelArg``` to Ocl specific include

runtime/runtime.h
bpervan: Same as for ```struct KernelArgs```, I have moved these two macros to the Ocl specific include

michel-steuwer (approved):

Issue #122, opened 2021-03-15
Make `Kind.Identifier` and `Kind.Explicitness` sealed traits


umazalakain:
Both `Kind.Identifier` and `Kind.Explicitness` are open traits, however we often rely on pattern matching on their implementation types.

Issue #103, opened 2021-02-11
Refactor DPIA traversals


umazalakain:
Refactor DPIA traversals following the refactoring of rise traversals (#98). DPIA traversals are already a lot closer to the current rise traversals. We could however benefit from generalising `Stop` and `Continue` to a monad, and invoking visitors directly (instead of returning `Continue(t, visitor)` and then invoking `visitor.visit(t)` we can just return visitor.visit(t)`).

PR #190, opened 2021-06-19
Do not extend `NamedVar` with `Kind.Identifier` in DPIA

umazalakain:
This way `NamedVar`s are uniform across RISE, DPIA, and arithexpr. This accomplishes two things:
- it is a first step towards the unification of types across RISE and DPIA
- it renders the set of bugs product of `DPIA -> arithexpr -> DPIA` translations irrelevant.

### DIFF
--- a/src/main/scala/shine/DPIA/Phrases/Phrase.scala
+++ b/src/main/scala/shine/DPIA/Phrases/Phrase.scala
@@ -38,21 +38,21 @@ final case class Apply[T1 <: PhraseType, T2 <: PhraseType](fun: Phrase[T1 ->: T2
   override def toString: String = s"($fun $arg)"
 }
 
-final case class DepLambda[T, I <: Kind.Identifier, U <: PhraseType](kind: Kind[T, I], x: I, body: Phrase[U])
-  extends Phrase[I `()->:` U] {
-  override val t: DepFunType[I, U] = DepFunType[I, U](kind, x, body.t)
-  override def toString: String = s"Λ(${x.name} : ${kind.name}). $body"
+final case class DepLambda[T, I, KI <: Kind.Identifier, U <: PhraseType](kind: Kind[T, I, KI], x: I, body: Phrase[U])
+  extends Phrase[DepFunType[I, KI, U]] {
+  override val t: DepFunType[I, KI, U] = DepFunType[I, KI, U](kind, x, body.t)
+  override def toString: String = s"Λ(${Kind.idName(kind, x)} : ${kind.name}). $body"
 }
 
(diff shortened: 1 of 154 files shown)


umazalakain:
@Bastacyclop @michel-steuwer This PR:
- renders #159 and #160 unnecessary
- enables #103 / #161
- solves #122

src/main/scala/shine/DPIA/Phrases/Phrase.scala
```diff
@@ -130,43 +130,10 @@ object Phrase {
...
-          }).asInstanceOf[N]
-
-          override def data[T <: DataType](dt: T): T = (dt match {
-            case i: DataTypeIdentifier =>
-              DataTypeIdentifier(idMap.getOrElse(i.name, i.name))
-            case dt => dt
-          }).asInstanceOf[T]
-        }
-        VisitAndRebuild(p, Renaming(Map()))
-      }
```
Bastacyclop, line 163: I think that if we don't try to enforce unique names anymore, we should use proper capture-avoiding substitution here?

src/main/scala/shine/OpenCL/Compilation/Passes/FlagPrivateArrayLoops.scala
```diff
@@ -79,7 +81,9 @@ object FlagPrivateArrayLoops {
                 pf.init, pf.n, pf.step, pf.dt, pf.out, pf.body), this)
             case _ => throw new Exception("This should not happen")
           }
-        case pf@ParForNat(level, dim, _, name) if (eliminateVars(pf.body.asInstanceOf[DepLambda[_, _ <: Kind.Identifier, _]].x.name)) =>
+        case pf@ParForNat(level, dim, _, name) if (eliminateVars(Kind.idName(
+          pf.body.asInstanceOf[DepLambda[_, _, _ <: Kind.Identifier, _]].kind,
+          pf.body.asInstanceOf[DepLambda[_, _, _ <: Kind.Identifier, _]].x))) =>
```
Bastacyclop, line 86: Maybe there is a better way to write this to avoid triple casting.

Bastacyclop (approved):

src/main/scala/shine/DPIA/Phrases/Phrase.scala
michel-steuwer: @Bastacyclop can you create an issue for that, please?

Bastacyclop: https://github.com/rise-lang/shine/issues/191

Issue #194, opened 2021-06-30
[BUG] Distribute constraints among tuning parameters creates dependency cycle 


johanneslenfers:
**Describe the bug**
Distributing constraints among tuning parameters creates dependency cycle. Feature is only necessary once the release Hypermapper version gets support for constraints in the config file. 
Method `distributeConstraints` in [configFileGeneration.scala](https://github.com/rise-lang/shine/blob/autotuning_merge/src/main/scala/rise/autotune/configFileGeneration.scala)

**To Reproduce**
Execute Test *distribute constraints* (line 441) in [autotuning.scala](https://github.com/rise-lang/shine/blob/autotuning_merge/src/test/scala/rise/autotuning.scala) 

**Expected behavior**
Output will show parameters, their constraints and dependencies. Parameter *tile* will be depended from *vec*. Parameter *vec* will be dependend from *tile*. Cycle.

PR #196, opened 2021-07-15
fixed distribution of dependencies among tuning parameters

johanneslenfers:
## Fixed issue #194
Distribute constraints among tuning parameters creating an implicit dependency graph. 

## Notes to Solution
- removed special case of `RangeConstraint`
- implicit dependency graph is now created without cycles
- added function to check a distribution for cycles
- automatic config file generation works again


src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -211,68 +211,34 @@ object configFileGeneration {
...
       Map[NatIdentifier, (Set[Constraint], Set[NatIdentifier])]()
-    parameters.foreach(param => {
+    parameters.toSeq.sortBy(_.name).foreach(param => {
       parametersWDC(param) = (Set.empty[Constraint], Set.empty[NatIdentifier])
     })
 
     // get parameters from constraint
     // check for given parameters in the given constraint
-    constraints.foreach(constraint => {
+    val test = constraints.toSeq.sortBy(_.toString)
```
Bastacyclop, line 220: is this used?

src/test/scala/rise/autotuning.scala
```diff
@@ -119,6 +119,37 @@ class autotuning extends test_util.Tests {
   }
   // scalastyle:on
 
+
+  // function to check cycle in parameter dependencies
+  def check_no_cycle(distribution: Map[NatIdentifier,
```
Bastacyclop, line 124: is this function only useful for testing, or should it be moved to `src/main`?

src/test/scala/rise/autotuning.scala
```diff
@@ -119,6 +119,37 @@ class autotuning extends test_util.Tests {
...
+              })
+            }
+          }
+        }
+      }
+    }
+
+    distribution.exists(param => {
+      check_no_cycle_rec(param._1, param._2._2, distribution)
+    })
```
Bastacyclop, line 150: I am not sure to understand your algorithm here. There is no cycle if there exists a parameter which does not induce a cycle? Would this example pass the check:
```
A depends on B
B depends on A
C depends on D
```
I am also surprised that you do not use a set of visited nodes/parameters in your algorithm.

src/main/scala/rise/autotune/configFileGeneration.scala
johanneslenfers: removed

src/test/scala/rise/autotuning.scala
johanneslenfers: I moved it to the config file generation of the auto tuning, where the parameter distribution is implemented.

src/test/scala/rise/autotuning.scala
johanneslenfers: You are right. This should not work properly. I updated the algorithm.

johanneslenfers: The algorithm is updated.

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -299,4 +300,35 @@ object configFileGeneration {
...
+                           dependencies: Set[NatIdentifier],
+                           distribution: Map[NatIdentifier,
+                             (Set[constraints.Constraint], Set[NatIdentifier])],
+                           visited: Set[NatIdentifier]
+                          ): Boolean = {
+
+      dependencies.size match {
+        case 0 => true
+        case _ => dependencies.exists(dependency => {
+          dependency.name.equals(param.name) || visited.contains(dependency)
```
Bastacyclop, line 318: You could avoid testing `dependency.name.equals(param.name)` by adding `param` to `visited` in the beginning.

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -299,4 +300,35 @@ object configFileGeneration {
...
+              distribution.apply(dependency)._2,
+              distribution,
+              visited + dependency)
+          })
+        }
+      }
+    }
+
+    distribution.forall(param =>
+      check_no_cycle_rec(param._1, param._2._2, distribution, Set.empty[NatIdentifier]))
```
Bastacyclop, line 332: Complexity-wise, I think that what you are doing is roughly `O(V*(V + E))`: for every `V` vertices/parameters, you might visit the entire graph of `V` parameters and `E` edges/dependencies. If you care about performance (and also want to use a time-tested algorithm), you can use standard `O(V + E)` algorithms:
https://www.geeksforgeeks.org/detect-cycle-in-a-graph/
or
https://www.geeksforgeeks.org/detect-cycle-direct-graph-using-colors/

Bastacyclop (approved):

PR #197, opened 2021-07-23
WIP: Add support for GAP8 hardware convolution engine

bpervan:
This PR will include everything necessary to utilize the hardware convolution engine featured by the GAP8 (convolution with filter of size 3x3, 5x5, 7x7, or 7x4)


src/main/scala/util/gen.scala
```diff
@@ -119,8 +120,8 @@ object gen {
                                           ): Expr => String =
         functionFromExpr(name, gen) andThen
           GAP8.Module.injectUnpacking andThen
-          C.Module.translateToString andThen
-          run(SyntaxChecker(_))
+          C.Module.translateToString// andThen
+          //run(SyntaxChecker(_))
```
michel-steuwer, line 124: Can this be commented back in? Or alternative, delete it completely.

michel-steuwer (approved):
Looks good and clear to me. 👍

src/main/scala/util/gen.scala
bpervan: Sure, I can't remember why I kicked that out in the first place. Anyways, I introduced it again with a hint that it should be done more thoroughly (this syntax checks only accelerator function + some declarations)

PR #198, opened 2021-07-28
Autotuning update

johanneslenfers:
## Update of the Auto Tuning
This PR includes improvements, bugfixes and features for the auto tuning. 

#### Refactoring
- Splitted auto tuning and auto tuning tests into multiple files.  

#### Improved config file generation: [here](https://github.com/rise-lang/shine/blob/autotuning_update/src/main/scala/rise/autotune/configFileGeneration.scala)
- filter out values based on statical-evaluable constraints
- inject input sizes into constraints 
- give input sizes as argument to the tuner, do not treat them as tuning parameters

#### Execution Iterations: [here](https://github.com/rise-lang/shine/blob/autotuning_update/src/main/scala/rise/autotune/execution.scala)
- execute generated kernel multiple times
- configure runtime measurement as Median or Minimum
- configure a `speedupFactor`, drop iterations, if execution has higher runtime than `speedupFactor * best` 

#### Collect information
- collect time to evaluate a sample (round trip) and its sub parts (codegen, compilation, execution)
- save tuning files with proper folder structure to disk 
- plot tuning results automatically

#### Failure Mode
- configure response to hypermapper in case of failure (-1 or intmax). 

### Bugfixes 
- deterministic constraints distribution: [here](https://github.com/rise-lang/shine/blob/autotuning_update/src/main/scala/rise/autotune/configFileGeneration.scala)


Bastacyclop (commented):
What are the included JSON files? Are they manually written, automatically generated? When do you manually write the JSON, and when do you use your JSON generator?

src/main/scala/rise/autotune/package.scala
```diff
@@ -19,19 +20,32 @@ import scala.sys.process._
...
                    hierarchicalHM: Boolean = false)
 
   case class Sample(parameters: Map[NatIdentifier, Nat],
                     runtime: Option[TimeSpan[Time.ms]],
                     timestamp: Long,
-                    autoTuningError: AutoTuningError)
+                    autoTuningError: AutoTuningError,
+                    tuningTimes: TuningTimes)
+
+  case class TuningTimes(roundTrip: Option[TimeSpan[Time.ms]],
```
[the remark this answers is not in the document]
Bastacyclop: is the name `total` not more adapted?

src/test/scala/rise/autotuning.scala
```diff
@@ -307,10 +318,149 @@ class autotuning extends test_util.Tests {
...
+    val n: Nat = 1024
+    val o: Nat = 1024
+    val map = Map(nIdent -> n, mIdent -> m, oIdent -> o)
+
+    val parameters = autotune.constraints.collectParameters(e)
+    val constraints = autotune.constraints.collectConstraints(e, parameters)
+    val constraintsSubstituted = constraints.
+      map(constraint => constraint.substitute(map.asInstanceOf[Map[ArithExpr, ArithExpr]]))
+
+    val dc = distributeConstraints(parameters, constraintsSubstituted)
```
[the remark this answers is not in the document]
johanneslenfers: The release version of Hypermapper used in the continuous integration does not support constraints, so we do not want to generate them here.

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -1,158 +1,181 @@
...
   def generateJSON(p: Parameters,
                    c: Set[Constraint],
-                   tuner: Tuner)
-  : String = {
+                   tuner: Tuner
+                  ): String = {
+
+    val parametersWDCImmutable = distributeConstraints(p, c)
+    val parametersWDC = scala.collection.mutable
+      .Map.empty[NatIdentifier, (Set[Constraint], Set[NatIdentifier])]
```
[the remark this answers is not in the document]
johanneslenfers: We remove the constraints from the map, which can be evaluated statically and therefore are used to filter out values during the config-file generation. Check [here](https://github.com/rise-lang/shine/blob/930f5678e02f4b9630d7e0bf3bbfb28ae11a8de8/src/main/scala/rise/autotune/configFileGeneration.scala#L117). I don't think this works with an immutable map, but maybe I missunderstood something about mutable/immutable here.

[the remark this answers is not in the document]
Bastacyclop: Can you not replace:
```scala
  parametersWDC.foreach(param => {
    // .. parametersWDC.apply(param._1) ..
    parametersWDC(param._1) = (constraintsFiltered, param._2._2)
    // .. parametersWDC(param._1) ..
```
With:
```scala
parametersWDCImmutable.foreach { case (param, wdc) => {
    // .. wdc ..
    val newWdc = (constraintsFiltered, param._2._2)
    // .. newWdc ..
```
?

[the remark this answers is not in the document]
johanneslenfers: You are right, the mutable Map is unnecessary. First, I didn't check what you meant. I applied the changes.

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -213,20 +165,72 @@ object configFileGeneration {
...
+      val constraintString = constraint match {
+        case RangeConstraint(n, r) => {
+          val (start, stop, step) = r match {
+            case RangeAdd(start, stop, step) => (start, stop, step)
+            case RangeMul(start, stop, step) => (start, stop, step)
+            case _ => (0, 0, 0) // todo catch other types of ranges
+          }
+
+          // if stop is PosInf, remove constraint
+          // (already catched by the range of parameter)
```
Bastacyclop, line 182: can you explain this? I am not sure I understand why removing the constraint is legal here.

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -213,20 +165,72 @@ object configFileGeneration {
...
+            case PosInf =>{
+              val startConstraint = n.toString + " >= " + start
+              val stepConstraint = n.toString + " % " + step + " == 0"
+
+              startConstraint + " and " + stepConstraint
+            }
+            case _ => {
+              val startConstraint = n.toString + " >= " + start
+              val stopConstraint = n.toString + " <= " + stop
+              val stepConstraint = n.toString + " % " + step + " == 0"
```
Bastacyclop, line 193: does this really work for `RangeMul`? (same above)

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -213,20 +165,72 @@ object configFileGeneration {
...
+              startConstraint + " and " + stopConstraint + " and " + stepConstraint
+            }
+          }
+
+        }
+        case PredicateConstraint(n) => {
+          n.toString.contains("/^") match {
+            case true => constraint.toString.replace("/^", "/")
+            case false => constraint.toString
+          }
```
Bastacyclop, line 204: why not simply `constraint.toString.replace("/^", "/")`?

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -213,20 +165,72 @@ object configFileGeneration {
...
+    list.size match {
+      case 0 => "[]"
+      case _ =>
+        var valuesString = ""
+        list.foreach(value => {
+          valuesString += value.toString +  ", "
+        })
+        "["  + valuesString.dropRight(2) + "]"
+    }
   }
```
Bastacyclop, line 222: `list.mkString("[", ", ", "]")`?

src/main/scala/rise/autotune/configFileGeneration.scala
```diff
@@ -213,20 +165,72 @@ object configFileGeneration {
...
-    "["  + valuesString.dropRight(2) + "]"
+    list.size match {
+      case 0 => "[]"
+      case _ =>
+        var valuesString = ""
+        list.foreach(value => {
+          valuesString += "\"" + value.toString + "\"" + ", "
+        })
+        "["  + valuesString.dropRight(2) + "]"
+    }
```
Bastacyclop, line 233: `mkString`?

src/main/scala/rise/autotune/configFileGeneration.scala
johanneslenfers: The constraints gives us the information that all values of the parameter have to be smaller than `PosInf`. We can drop this information, because it is a requirement for Hypermapper to work. Otherwise, it would not possible to generate the search space. If `stop` has a certain - finitie - value, we keep this information.

Bastacyclop: Would it not make more sense to replace `PosInf` with `Int.MaxValue`? Or is this what Hypermapper does?

Bastacyclop: PS: I think the comment "if stop is PosInf, remove constraint" is misleading here, I thought you meant removing the entire range constraint including start and steps bits.

src/main/scala/rise/autotune/execution.scala
```diff
@@ -3,75 +3,207 @@ package rise.autotune
...
+
+case class ExecutionResult(runtime: Option[TimeSpan[Time.ms]],
+                           error: AutoTuningError,
+                           codegenTime: Option[TimeSpan[Time.ms]],
+                           compilationTime: Option[TimeSpan[Time.ms]],
+                           executionTime: Option[TimeSpan[Time.ms]],
+                          )
+
 object execution {
+  var best:Option[Double] = None
```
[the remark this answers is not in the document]
Bastacyclop: ping

Bastacyclop (approved):
Looks good to me overall, I didn't check the details in `rise.autotune` since I have done this previously. Is there any outstanding issue to document here? Or things that don't work?

PR #200, opened 2021-08-11
Mixed work on scaling equality saturation to tiling, mm, etc

Bastacyclop:
Includes:
- tests and benchmarks for reordering and tiling
- code for direct e-class substitution, but the code is broken and not maintained
  - should this be merged in or put on the side?
- hash-consing for nats and types
- basic e-class filtering algorithm
- `EGraphDot` can now only print e-classes which are reachable from a set of given ones
- extraction of random terms from an e-graph
- hash-consing for pattern substitutions (can save a lot of RAM usage but does not help CPU usage as-is)
  - should this be merged in or put on the side? it should at least be optional
- `NotFreeIn` named rewrite rule conditional
- `ProveEquiv` builder
- directed rewriting prototype implementation, but it is not correct in the general case
  - should this be merged in or put on the side?
- rewrite rule `Scheduler` similar to egg`s design
- e-graph `TypeCheck` procedure (does not check everything but can catch bugs)
- many rewrite-rules, some of them might be questionable
- experiments with `>>` function composition combinator
- examples of beta-reduction and eta-reduction subtleties
- initial experiments rewriting MM and Harris applications


src/main/scala/rise/eqsat/Analysis.scala
```diff
@@ -8,57 +8,212 @@ case object Greater extends Order
 /** Explains how arbitrary analysis data associated with an [[EClass]]
   * is maintained across [[EGraph]] operations.
   * @see [[https://docs.rs/egg/0.6.0/egg/trait.Analysis.html]]
+  * @todo try to reduce the generics boilerplate with Scala 3 features
```
michel-steuwer, line 11: Hopefully we will be on Scala 3 soon ... :-)

src/main/scala/rise/eqsat/EGraph.scala
```diff
@@ -1,28 +1,39 @@
...
+      hashConses = HashConses.emptyWithAnalysis(analysis),
+    )
 }
 
 /** A data structure to keep track of equalities between expressions.
   * @see [[https://docs.rs/egg/0.6.0/egg/struct.EGraph.html]]
   */
-class EGraph[Data](
-  val analysis: Analysis[Data],
+class EGraph[ED, ND, TD](
```
michel-steuwer, line 24: Can you maybe add a comment to indicate what `ED`, `ND`, and `TD` are (I assume something like Expr Data, Nat Data, Type Data???)

michel-steuwer (approved):
I have not looked at every line in detail (and there seems to be plenty of things in flux), but I am happy for this to be merged (once you fix the merge conflicts and the tests all pass).

src/main/scala/rise/eqsat/EGraph.scala
Bastacyclop: done

michel-steuwer (approved):

PR #201, opened 2021-08-13
Preparation for Scala 3 migration

michel-steuwer:
This PR introduces changes that will help to ease the migration to Scala 3, but the code in this PR remains valid Scala 2 code.

In detail the changes are:
 - The type `()` is replaced with `Unit` (as Scala 3 seems to no longer accept `()` as a type)
 - The use of generic type projections (i.e. `T#U`) in the `FunctionHelper` class is avoided as these have been [dropped](https://docs.scala-lang.org/scala3/reference/dropped-features/type-projection.html) in Scala 3
 - The `._1` and `._2` notations on Rise and DPIA expressions do not work in Scala 3 (I don't know why) I have replaced them with `` `1` `` and `` `2` ``
 - The `` `_` `` notation is also not accepted by Scala 3, I have replaced it with `` `__` ``
 - Returning anonymous object types (i.e. `def foo() = new Object { ... } has changed somehow and I have avoided this now altogether by giving the returned type names. This was mainly used in the DSLs.
 - I accommodated a [change](https://docs.scala-lang.org/scala3/guides/migration/incompat-other-changes.html#explicit-call-to-unapply) in the signature of the `unapply` method generated for case classes.
 - I removed some syntax that used Unicode symbols. This is not related to Scala 3 but has repeatedly led to problems to compile on systems where the file system does not support Unicode (as Scala generates a file with the class name for each class).

### DIFF
--- a/src/main/scala/apps/convolution.scala
+++ b/src/main/scala/apps/convolution.scala
@@ -23,8 +23,8 @@ object convolution {
 
   private val dotElemWeightsSeq = fun((weights, elem) =>
     oclReduceSeqUnroll(AddressSpace.Private)(fun((acc, pair) => {
-      val pixel = pair._1
-      val weight = pair._2
+      val pixel = pair.`1`
+      val weight = pair.`2`
       acc + (pixel * weight)
     }))(lf32(0.0f))(zip(join(elem))(weights)))
 
@@ -143,9 +143,7 @@ object convolution {
     matrix: Array[Array[Float]],
     weights: Array[Float]
   ): (Array[Float], TimeSpan[Time.ms]) = {
-    val f = k.as[ScalaFunction `(`
-      Array[Array[Float]] `,` Array[Float]
-      `)=>` Array[Float]]
+    val f = k.as[Args `(` Array[Array[Float]] `,` Array[Float], Array[Float]]
     f(localSize, globalSize)(matrix `,` weights)
   }
 }
--- a/src/main/scala/rise/core/IsClosedForm.scala
+++ b/src/main/scala/rise/core/IsClosedForm.scala
@@ -69,25 +69,37 @@ object IsClosedForm {
 
     override def natToData: NatToData => Pair[NatToData] = {
       case NatToDataLambda(x, e) =>
-        for { p <- this.copy(boundT = boundT + NatKind.IDWrapper(x)).`type`(e)}
-          yield (p._1, NatToDataLambda(x, e))
+//        for { p <- this.copy(boundT = boundT + NatKind.IDWrapper(x)).`type`(e)}
+//          yield (p._1, NatToDataLambda(x, e))
+        val pp: Pure[((OrderedSet[Identifier], OrderedSet[Kind.Identifier]), DataType)] =
+          this.copy(boundT = boundT + NatKind.IDWrapper(x)).`type`(e)
+        pp.map(p => (p._1, NatToDataLambda(x, e)))
       case t => super.natToData(t)
     }
 
     override def natToNat: NatToNat => Pair[NatToNat] = {
       case NatToNatLambda(x, n) =>
-        for { p <- this.copy(boundT = boundT + NatKind.IDWrapper(x)).nat(n)}
-          yield (p._1, NatToNatLambda(x, n))
+//        for { p <- this.copy(boundT = boundT + NatKind.IDWrapper(x)).nat(n)}
+//          yield (p._1, NatToNatLambda(x, n))
+        val pp: Pure[((OrderedSet[Identifier], OrderedSet[Kind.Identifier]), Nat)] =
+          this.copy(boundT = boundT + NatKind.IDWrapper(x)).nat(n)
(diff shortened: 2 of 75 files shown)


src/main/scala/apps/convolution.scala
```diff
@@ -143,9 +143,7 @@ object convolution {
     matrix: Array[Array[Float]],
     weights: Array[Float]
   ): (Array[Float], TimeSpan[Time.ms]) = {
-    val f = k.as[ScalaFunction `(`
-      Array[Array[Float]] `,` Array[Float]
-      `)=>` Array[Float]]
+    val f = k.as[Args `(` Array[Array[Float]] `,` Array[Float], Array[Float]]
```
Bastacyclop, line 146: I find this syntax very confusing, especially with the missing `)`. Can we have something like:
```scala
k.as[`In:` Array[Array[Float]] `,` Array[Float], `Out:` Array[Float]]
```

src/main/scala/rise/core/IsClosedForm.scala
```diff
@@ -69,25 +69,37 @@ object IsClosedForm {
 
     override def natToData: NatToData => Pair[NatToData] = {
       case NatToDataLambda(x, e) =>
-        for { p <- this.copy(boundT = boundT + NatKind.IDWrapper(x)).`type`(e)}
-          yield (p._1, NatToDataLambda(x, e))
+//        for { p <- this.copy(boundT = boundT + NatKind.IDWrapper(x)).`type`(e)}
+//          yield (p._1, NatToDataLambda(x, e))
+        val pp: Pure[((OrderedSet[Identifier], OrderedSet[Kind.Identifier]), DataType)] =
+          this.copy(boundT = boundT + NatKind.IDWrapper(x)).`type`(e)
+        pp.map(p => (p._1, NatToDataLambda(x, e)))
```
Bastacyclop, line 76: is the `for` syntax not supported anymore?

src/main/scala/rise/eqsat/NamedRewrite.scala
```diff
@@ -414,7 +414,7 @@ object NamedRewriteDSL {
 
   import scala.language.implicitConversions
 
-  val `_`: rct.TypePlaceholder.type = rct.TypePlaceholder
+  val `__`: rct.TypePlaceholder.type = rct.TypePlaceholder
```
Bastacyclop, line 417: maybe we can also just remove that syntax at that point and always use names ..

src/main/scala/shine/DPIA/Phrases/Phrase.scala
```diff
@@ -47,9 +47,9 @@ final case class DepLambda[T, I, U <: PhraseType](kind: Kind[T, I],
...
 object DepLambda {
-  def apply[T, I](kind: Kind[T, I], x: I): Object {
-    def apply[U <: PhraseType](body: Phrase[U]): DepLambda[T, I, U]
-  } = new {
+  def apply[T, I](kind: Kind[T, I], x: I): SyntaxHelper[T, I] = SyntaxHelper(kind, x)
+
+  case class SyntaxHelper[T, I](kind: Kind[T, I], x: I) {
     def apply[U <: PhraseType](body: Phrase[U]): DepLambda[T, I, U] = DepLambda(kind, x, body)
   }
 }
```
Bastacyclop, line 55: Can we use a `case class DepLambda` to avoid the helper?

src/main/scala/apps/convolution.scala
michel-steuwer: Don't obsess over this, I know already how to fix this in a much nicer way in Scala 3 using [match types](https://docs.scala-lang.org/scala3/reference/new-types/match-types.html):
https://scastie.scala-lang.org/ndV4Xu8RQZWlZRYCc6GAig

src/main/scala/rise/core/IsClosedForm.scala
michel-steuwer: No, of course it is, but here the system got confused somehow (and I with it ...).
I think that maybe the type inference has changed, and therefore the type annotation helps here, but I am not sure.
The explicit version with the `map` should be the correct one - from my point of view at least.

src/main/scala/rise/eqsat/NamedRewrite.scala
michel-steuwer: That is completely up to you, I just renamed it in a way that Scala 3 does not compain about ...

src/main/scala/shine/DPIA/Phrases/Phrase.scala
michel-steuwer: Of course, there is a class `DepLambda` that represents a dependent lambda phrase.
We usually keep those classes free of any syntactic sugar (which I think is a good idea) and even for this here it is not clear if this syntax should be defined here or elsewhere in a DSL package.

Anyway, I think the purpose of this is only to infer the type `U` and not have to annotate this, while still allowing the annotation of `T` and `I`.

We might want to look into this after having migrated to Scala 3 again, I supose ...

PR #203, opened 2021-08-19
Fix bug in HostManagedBuffers logic

Bastacyclop:
This didn't work, hitting an assertion:
```
depFun((n: Nat) => fun((n`.`i32) ->: (n`.`i32))(in =>
  mapSeq(add(li32(2)))(in) |> store(x =>
  oclRun(LocalSize(2), GlobalSize(n/2))(mapGlobal(add(li32(1)))(x)))
))
```
Now, it generates:
```c
#include "ocl/ocl.h"
struct foo_t {
  Kernel k0;
};

typedef struct foo_t foo_t;

void foo_init(Context ctx, foo_t* self){
  (*self).k0 = loadKernel(ctx, k0);
}

void foo_destroy(Context ctx, foo_t* self){
  destroyKernel(ctx, (*self).k0);
}

void foo_run(Context ctx, foo_t* self, Buffer moutput, int n1, Buffer me2){
  {
    Buffer mx101 = createBuffer(ctx, n1 * sizeof(int32_t), HOST_WRITE | DEVICE_READ);
    {
      int32_t* x101 = (int32_t*)hostBufferSync(ctx, mx101, n1 * sizeof(int32_t), HOST_WRITE);
      int32_t* e2 = (int32_t*)hostBufferSync(ctx, me2, n1 * sizeof(int32_t), HOST_READ);
      /* mapSeq */
      for (int i_110 = 0; i_110 < n1; i_110 = 1 + i_110) {
        x101[i_110] = ((int32_t)2) + e2[i_110];
      }
      
    }
    
    {
      DeviceBuffer b0 = deviceBufferSync(ctx, moutput, n1 * sizeof(int32_t), DEVICE_WRITE);
      DeviceBuffer b2 = deviceBufferSync(ctx, mx101, n1 * sizeof(int32_t), DEVICE_READ);
      const size_t global_size[3] = (const size_t[3]){n1 / 2, 1, 1};
      const size_t local_size[3] = (const size_t[3]){2, 1, 1};
      const KernelArg args[3] = (const KernelArg[3]){KARG(b0), KARG(n1), KARG(b2)};
      launchKernel(ctx, (*self).k0, global_size, local_size, 3, args);
    }
    
    destroyBuffer(ctx, mx101);
  }
  
}

void foo_init_run(Context ctx, Buffer moutput, int n1, Buffer me2){
  foo_t foo;
  foo_init(ctx, &foo);
  foo_run(ctx, &foo, moutput, n1, me2);
  foo_destroy(ctx, &foo);
}
```
However there seems to be another type of bug remaining. It looks like the "one_copy" buffer implementation is failing to produce the correct output.

### DIFF
--- a/src/main/scala/shine/DPIA/showDPIA.scala
+++ b/src/main/scala/shine/DPIA/showDPIA.scala
@@ -75,6 +75,8 @@ case class showDPIA(showTypes : ShowTypes = Off) {
     case Cast(_, dt2, e) => s"cast ${showKindedType(DataKind, dt2)} ${showPhrase(e)}"
     case RotateValues(_,_,_,wrt,in) => s"rotate ${showPhrase(in)}\n${tab(showPhrase(wrt))}"
     case f : ReduceSeq => s"reduce ${showPhrase(f.array)} ${showPhrase(f.init)}\n${tab(showPhrase(f.f))}"
+    case kc: shine.OpenCL.primitives.imperative.KernelCallCmd =>
+      s"${kc.name}<<<${kc.localSize}, ${kc.globalSize}>>>(${kc.args}, ${kc.output})"
     case p => p.toString
   }
 
--- a/src/main/scala/shine/DPIA/showDPIA.scala
+++ b/src/main/scala/shine/DPIA/showDPIA.scala
@@ -47,7 +47,7 @@ case class showDPIA(showTypes : ShowTypes = Off) {
   def showPrim : Primitive[_] => String = {
     case Comment(s) => s"/* ${s} */"
     case Seq(c1, c2) => s"${showPhrase(c1)};\n${showPhrase(c2)}"
-    case New(dt, f) => s"new (exp[${showDataType(dt)}] x acc[${showDataType(dt)}]) in \n${tab(showPhrase(f))}"
+    case New(dt, f) => s"new (${showType(f.t.inT)}) in \n${tab(showPhrase(f))}"
     case Assign(_, lhs, rhs) => s"${showTypedPhrase(lhs)} = ${showPhrase(rhs)}"
     case IdxAcc(_, _, idx, array) => s"${showTypedPhrase(array)}[${showPhrase(idx)}]"
     case Idx(_, _, idx, array) => s"${showTypedPhrase(array)}[${showPhrase(idx)}]"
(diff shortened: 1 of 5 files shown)


fedepiz:
Seems to be working on my side, at least the small tests, and it allows me to generate the multi-kernel scan. Not sure about the remaining bug because I am yet to actually run that code - but it's ok to merge this in for now, it's still a step forward.

fedepiz (approved):
Seems to be working on my side, at least the small tests, and it allows me to generate the multi-kernel scan. Not sure about the remaining bug because I am yet to actually run that code - but it's ok to merge this in for now, it's still a step forward.

src/main/scala/shine/DPIA/showDPIA.scala
```diff
@@ -75,6 +75,8 @@ case class showDPIA(showTypes : ShowTypes = Off) {
     case Cast(_, dt2, e) => s"cast ${showKindedType(DataKind, dt2)} ${showPhrase(e)}"
     case RotateValues(_,_,_,wrt,in) => s"rotate ${showPhrase(in)}\n${tab(showPhrase(wrt))}"
     case f : ReduceSeq => s"reduce ${showPhrase(f.array)} ${showPhrase(f.init)}\n${tab(showPhrase(f.f))}"
+    case kc: shine.OpenCL.primitives.imperative.KernelCallCmd =>
+      s"${kc.name}<<<${kc.localSize}, ${kc.globalSize}>>>(${kc.args}, ${kc.output})"
```
michel-steuwer, line 79: Is this using CUDA syntax for the OpenCL kernel call?

Can we maybe avoid this (possible confusion) even when this is just a debug output?

michel-steuwer (dismissed):
It doesn't seem that the remaining bug is conceptual, as in that a `*BufferSync` sync is missing (or is it?).
If that's not the case, then that suggests that the "one_copy" implementation gets confused over its internal flags?

Fine for me to merge this, maybe after having addressed my one comment.

src/main/scala/shine/DPIA/showDPIA.scala
Bastacyclop: I was not sure what other syntax to use, I don't think that there is an OpenCL equivalent for this (it is higher level than `clEnqueueNDRangeKernel`). Or is there a SYCL equivalent?

michel-steuwer: I don't think there is a convenient SYCL syntax either. I just think that using the CUDA syntax can be quite confusing. How about just writing something like:
```
s"oclKernelLaunch(${kc.name})(${kc.localSize}, ${kc.globalSize})(${kc.args}, ${kc.output})"
```

Bastacyclop: I've done something like that and added some more OpenCL print cases as well. Here is an example:
```
// before HostManagedBuffers.insert
new (exp[n1.i32, read] x acc[n1.i32]) in 
  λ(x101).
    (/* mapSeq */;
    for n1
      λ(x109).
        (((proj2 (x101))[x109]) = (cast (i32) 2) + ((e2)[x109]));
    output = oclLaunchKernel(k0, LocalSize(2,1,1), GlobalSize(((n1) / (2)),1,1))(NatAsIntData(n1), proj1 (x101)))

// after analyzeAndInsertHostExecution
new (exp[n1.i32, read] x acc[n1.i32]) in 
  λ(x101).
    (oclHostExecution (e2 HOST_READ, x101 HOST_WRITE) in
      /* mapSeq */;
      for n1
        λ(x109).
          (((proj2 (x101))[x109]) = (cast (i32) 2) + ((e2)[x109]));
    output = oclLaunchKernel(k0, LocalSize(2,1,1), GlobalSize(((n1) / (2)),1,1))(NatAsIntData(n1), proj1 (x101)))

// after insertManagedBuffers
newManagedBuffer (exp[managed[n1.i32], read] x acc[managed[n1.i32]] HOST_WRITE | DEVICE_READ) in 
  λ(mx101).
    (oclHostExecution (e2 HOST_READ, x101 HOST_WRITE) in
      /* mapSeq */;
      for n1
        λ(x109).
          (((proj2 (x101))[x109]) = (cast (i32) 2) + ((e2)[x109]));
    moutput = oclLaunchKernel(k0, LocalSize(2,1,1), GlobalSize(((n1) / (2)),1,1))(NatAsIntData(n1), proj1 (mx101)))
```

michel-steuwer (approved):

PR #205, opened 2021-08-20
Scala 3 Draft PR

michel-steuwer:
This is a draft PR porting `shine` and all dependent repositories to Scala 3.

At the moment of opening this PR, there exists a large performance regression that me need to address before merging.

### DIFF
--- a/src/main/scala/apps/harrisCornerDetectionHalideRewrite.scala
+++ b/src/main/scala/apps/harrisCornerDetectionHalideRewrite.scala
@@ -32,16 +32,16 @@ object harrisCornerDetectionHalideRewrite {
     })
   }
 
-  val unrollDots: Strategy[Rise] = normalize.apply(lowering.reduceSeqUnroll)
+  val unrollDots: Strategy[Rise] = normalize(lowering.reduceSeqUnroll)
 
   def someGentleReduction: Strategy[Rise] =
     gentleBetaReduction() <+ etaReduction() <+
     idxReduction <+ fstReduction <+ sndReduction <+
     removeTransposePair
-  def reducedFusedForm: Strategy[Rise] = normalize.apply(
+  def reducedFusedForm: Strategy[Rise] = normalize(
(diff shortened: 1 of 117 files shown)


Bastacyclop:
I see that you have changed some lazy vals to defs, could that be part of the performance regression?

Bastacyclop:
In elevate macros you are wrapping these everywhere (strategies, combinators):
```scala
       try ... catch
          case _: MatchError => Failure(this)
```
I don't think this is what we used to do?
Ideally `Success`/`Failure` should only be automatically wrapped around rules and not strategies/combinators?

Bastacyclop:
In src/main/scala/elevate/core/strategies/debug.scala, there are rule definitions which look more like strategies to me. I think we should be careful about what is a rule and what is a strategy?

src/main/scala/apps/harrisCornerDetectionHalideRewrite.scala
```diff
@@ -279,7 +279,7 @@ object harrisCornerDetectionHalideRewrite {
           topDown(slideBeforeMapMapF)
         ) `;`
         reducedFusedForm `;`
-        normalize.apply(
+        normalize(
```
Bastacyclop, line 282: finally :)

src/test/scala/apps/convolution1D.scala
```diff
@@ -1,18 +1,20 @@
 package apps
 
 import apps.separableConvolution2D._
```
Bastacyclop, line 3: `import apps.separableConvolution2D.*`, or delete?

src/test/scala/apps/mmTensorCheck.scala
```diff
@@ -6,6 +6,8 @@ import shine.OpenCL._
 import shine.cuda.KernelExecutor.{KernelNoSizes, KernelWithSizes}
 import util._
```
Bastacyclop, line 7: what is the difference between `import x._` and `import x.*`?

PR #207, opened 2021-09-02
CGO18 Benchmarks

Bastacyclop:
Includes high-level expressions for CGO18's Figure 8 benchmarks.
- I have not checked that they are functionally correct
- Porting low-level expressions requires more work (e.g. from these folders https://gitlab.com/larisa.stoltzfus/liftstencil-cgo2018-artifact/-/tree/master/benchmarks/figure8/workflow1/gaussian/big/kepler)
- I only ported one expression from Figure 7 as low-level expressions are not included in the artifact, they would need to be re-regenerated or inferred from the generated OpenCL
- There is a good amount of code duplication which could be factored out if desired

@michel-steuwer should  we merge this?


michel-steuwer (approved):
👍

PR #216, opened 2021-09-24
Removed intermediate primitives

michel-steuwer:
This PR removes the intermediate primitives.

These added only an additional abstraction that has (in my experience) lead to more confusion that clarification when understanding the compilation process.

As (most) intermediate primitives were only ever used once, there is no reduction in code reuse.


src/main/scala/shine/DPIA/Compilation/AcceptorTranslation.scala
```diff
@@ -142,8 +151,8 @@ object AcceptorTranslation {
         (expT(dt1, read) ->: (comm: CommType)) ->: (comm: CommType)
       )(next =>
         comment("iterateStream") `;`
-          forNat(n, i =>
-            streamNext(next, i, fun(expT(dt1, read))(x => fI(x)(A `@` i))))
+          forNat(n = n, f = i =>
```
Bastacyclop, line 154: `name =`?

src/main/scala/shine/DPIA/Compilation/AcceptorTranslation.scala
```diff
@@ -223,10 +231,13 @@ object AcceptorTranslation {
...
       con(array)(λ(expT(n`.`dt1, read))(x =>
         con(init)(λ(expT(dt2, read))(y =>
-          ScanSeqI(n, dt1, dt2,
-            λ(expT(dt1, read))(x => λ(expT(dt2, read))(y => λ(accT(dt2))(o =>
-              acc(f(x)(y))(o)))),
-            y, x, A)))))
+          comment("scanSeq")`;`
+          `new`(dt2, accumulator =>
+            acc(y)(accumulator.wr) `;`
+            `for`(n = n, f = i =>
```
Bastacyclop, line 237: `name =`?

src/main/scala/shine/DPIA/Compilation/AcceptorTranslation.scala
```diff
@@ -274,18 +281,36 @@ object AcceptorTranslation {
...
-          _Λ_(NatKind)(l => λ(accT(l`.`dt))(o =>
-            λ(expT({l * n}`.`dt, read))(x => acc(f(l)(x))(o)))),
-          x)))
+      con(array)(λ(expT({m * n.pow(k)}`.`dt, read))(x => {
+        import arithexpr.arithmetic.Cst
+        val sz = n.pow(k) * m
+
+        shine.OpenCL.DSL.newDoubleBuffer(a, sz`.`dt, m`.`dt, sz`.`dt, x, A,
+          (v, swap, done) => {
+            shine.DPIA.DSL.`for`(n = k, f = ip => {
```
Bastacyclop, line 304: `name = `?

src/main/scala/shine/DPIA/Compilation/ContinuationTranslation.scala
```diff
@@ -62,8 +59,7 @@ object ContinuationTranslation {
           `if`(x) `then` con(thenP)(C) `else` con(elseP)(C)
         })
 
-      case Proj1(_) => throw new Exception("This should never happen")
-      case Proj2(_) => throw new Exception("This should never happen")
+      case Proj1(_) | Proj2(_) => C(E)
```
Bastacyclop, line 62: Where is this change coming from?

src/main/scala/shine/DPIA/Compilation/StreamTranslation.scala
```diff
@@ -43,15 +42,43 @@ object StreamTranslation {
   def primitive(E: ExpPrimitive)
                (C: Phrase[`(nat)->:`[(ExpType ->: CommType) ->: CommType] ->: CommType])
                (implicit context: TranslationContext): Phrase[CommType] = E match {
-    case CircularBuffer(n, alloc, sz, dt1, dt2, load, input) =>
+    case CircularBuffer(n, _, size, dt1, dt2, load, input) =>
```
Bastacyclop, line 45: Maybe you can add `assert(alloc == size)`.

src/main/scala/shine/DPIA/Compilation/StreamTranslation.scala
```diff
@@ -43,15 +42,43 @@ object StreamTranslation {
...
+                Zip(size, a, b, `read`, ra, rb))
+            ))
+          case _ =>
+            `new`(size `.` dt, buffer => C(buffer.wr, buffer.rd))
+        }
+
+        allocGen(dt2, (bufWr: Phrase[AccType], bufRd: Phrase[ExpType]) => {
+          // TODO: unroll flags?
+          // prologue initialisation
+          forNat(n = size - 1, f = i =>
```
Bastacyclop, line 65: `name =`?

src/main/scala/shine/DPIA/Compilation/StreamTranslation.scala
```diff
@@ -93,25 +137,81 @@ object StreamTranslation {
...
+                    Cycle(i + size, alloc, dt2, bufRd)))
+              ),
+              arithexpr.arithmetic.RangeAdd(0, n, 1)
+            ))
+        })
+      }
+//        oclI.CircularBufferI(a, n, alloc, sz, dt1, dt2,
+//          fun(expT(dt1, read))(x =>
+//            fun(accT(dt2))(o => acc(load(x))(o))),
+//          nextIn, C)
```
Bastacyclop, line 186: can be deleted?

Bastacyclop (commented):
Looks good to me, I added small comments. My only concern is that I believe the way you inlined some functions changes the translation evaluation order: before some variables would be translated as black-boxes before knowing their value, now their value is directly given to the translation. Do the changes have any impact on compilation time or generated code?

src/main/scala/rise/autotune/package.scala
```diff
@@ -35,7 +35,7 @@ package object autotune {
 
   // todo add meta information (configuration, times, samples, ...)
   case class TuningResult(samples: Seq[Sample])
-  case class AutoTuningError(errorLevel: AutoTuningErrorLevel, message: Option[String])
+//  case class AutoTuningError(errorLevel: AutoTuningErrorLevel, message: Option[String])
```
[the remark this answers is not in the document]
michel-steuwer: The `AutoTuningError` class exists twice. Here and in another file. This confuses my `sbt` setup (but obviously not everywhere ...). I will delete it from here altogether.

src/main/scala/shine/DPIA/Compilation/AcceptorTranslation.scala
```diff
@@ -129,11 +126,23 @@ object AcceptorTranslation {
...
-            λ(expT((l * n)`.`dt, read))(x => acc(f(l)(x))(o)))),
-          x)))
+      con(array)(λ(expT((m * n.pow(k))`.`dt, read))(x => {
+        val sz = n.pow(k) * m
+
+        newDoubleBuffer(sz`.`dt, m`.`dt, sz`.`dt, x, A,
+          (v: Phrase[VarType],
+           swap: Phrase[CommType],
+           done: Phrase[CommType]) => {
+            `for`(n = k, f = ip => {
```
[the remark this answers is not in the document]
michel-steuwer: No. I agree. I will simplify.

src/main/scala/shine/DPIA/Compilation/ContinuationTranslation.scala
michel-steuwer: As observed by you at the top, the evaluation order has changed now and this here is to allow the variables, i.e. (`accum.rd` and `accum.wr`) to work.

src/main/scala/shine/DPIA/Compilation/StreamTranslation.scala
michel-steuwer: > Looks good to me, I added small comments. My only concern is that I believe the way you inlined some functions changes the translation evaluation order: before some variables would be translated as black-boxes before knowing their value, now their value is directly given to the translation. Do the changes have any impact on compilation time or generated code? It seems like this change already required adding a translation for `IdxDistribute` for example (or does this come from some other change?).

That is right. This changes the evaluation order. But this could easily be restored without the intermediate patterns, just be re-adding the DPIA functions that added the indirection.

I think that these functions were unnecessary, and we should always have supported `IdxDistribute` and  `Fst` + `Snd`. In fact, we should try to get to a design where we have exhaustive checks over the primitives as well ...

I'll fix the many small issues that you found. I should have made a cleanup path before opening the PR. Thanks for spotting them.

src/main/scala/shine/DPIA/Compilation/ContinuationTranslation.scala
Bastacyclop: I thought we had a reason not to have a continuation translation for these .. but I am not sure I remember why. Maybe it is because such projections should only be introduced by the translation, and not exist before it? I guess now the translation ends up introducing projections and throwing them back at its own recursion, which did not happen before.

Bastacyclop (approved):

Issue #212, opened 2021-09-06
Unnecessary end-of-loop barriers


Bastacyclop:
At the end of a loop, there are two possible scenarios:
1. the loop will be taken again
2. the loop will be exited

Currently barrier insertion conflates the two, but sometimes more precise barriers could be inserted for the individual scenarios.
This is easy noticed for example on 1-trip loops such as in CGO17's Convolution Row benchmark.
We know that the loop is only taken once, but we stil insert a barrier for scenario 1. at the end of the loop.

Bastacyclop:
related to #18

PR #80, opened 2020-11-25
Work on barrier elimination

Bastacyclop:
This PR aims to improve on https://github.com/rise-lang/shine/issues/18, but will not solve the issue entirely.

PR #220, opened 2022-02-03
Port barrier tests from Lift

Bastacyclop:
This ports more than 40 barrier unit tests from the Lift repository to ours.
Overall, Lift seems to generate less barriers, but also to generate more invalid barriers (see point 2.). The barrier insertion of Shine is more conservative and often looks more correct to me, but could be improved for performance (see point 1.).
1. In many tests, Lift is able to generate less barriers. For Shine to do the same, the dependency analysis for barrier insertion needs to be more precise, currently it is very coarse-grain.
1. In many tests, Lift generates invalid barriers (e.g. using the wrong address space flag, or forgetting dependencies between loop round-trips), but the code seems to work by chance.
1. Some tests in Lift inject work sizes to ensure that some loops are only taken once, I ported them over without injecting the sizes. This should be corrected when fixing #212 
1. Some generated barriers might not be reached by all threads (already mentioned in #80). This needs to be solved at some point, Lift had a detection system to throw errors in such cases, but I think we could do better by automatically transforming the invalid code into valid code.
1. Some tests from Lift create writing races with patterns such as `mapSeq`/`reduceSeq` which are outside of a `mapLocal`. I think such programs should be rejected and considered ill formed.
1. In many tests, Lift generates less barriers by using more memory allocation, while Rise generates more barriers at the benefit of using less memory allocation. This is a hard-wired trade-off in both frameworks.
1. Some tests have invalid access annotations when ported to Rise, this is an orthogonal problem but deserves more thought.
1. Some tests generate record of arrays which breaks our code generation, another orthogonal problem.


johanneslenfers:
Porting the barrier tests from Lift makes sense to me. 

Just for my understanding: Is point 2 captured by the tests? So in Lift the tests would pass because the code works? Do these tests pass for Rise? Would it make sense to check for invalid barriers rather than working code? 

Are you going to keep track of these points as issue(s)?

johanneslenfers (approved):

PR #229, opened 2022-02-17
added missing assert to mm execution test in auto tuning

johanneslenfers:
The mm example execution as part of the auto tuning tests is a heavy workload if running on a non-gpu device. Therefore, the execution is allowed to fail with an `EXECUTION_ERROR`. This PR adds the missing `assert`.

### DIFF
--- a/src/test/scala/rise/autotune/TestExecution.scala
+++ b/src/test/scala/rise/autotune/TestExecution.scala
@@ -110,7 +110,7 @@ class TestExecution extends test_util.Tests {
 
       // check for execution error
       result.runtime match {
-        case Left(value) => value.errorLevel.equals(EXECUTION_ERROR)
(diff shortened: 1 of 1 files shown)


Bastacyclop:
I think ideally tests should be able to different execution errors. For example in this case I think the test should also be allowed to pass if the `CL_OUT_OF_RESOURCES` error is triggered?

The next step after that would be to query things like `clGetDeviceInfo(device, CL_DEVICE_MAX_WORK_ITEM_DIMENSIONS, ...)`, `CL_DEVICE_MAX_WORK_ITEM_SIZES`, `CL_DEVICE_MAX_WORK_GROUP_SIZES`, `CL_DEVICE_LOCAL_MEM_SIZE`, ...
And either check them against `clGetKernelWorkGroupInfo` data, or even better allow this info to drive autotuning (as was done in Lift I believe, at least to some extent).

PR #230, opened 2022-04-25
Experiment with host code generation on old GPGPU 2020 benchmarks

Bastacyclop:
Previously I wrote about 1000 lines of host code by hand, now we can generate host code automatically instead (about 1200 lines).
1. I did not try running the code yet
1. I had to implement some hack to counter the fact that two unified nats may not have equal normal forms after type inference (the problem was there before and is completely orthogonal)
1. I had to replace global temporary memory with local temporary memory, as support for global temporary memory has not yet been implemented
1. I had to convert vectorized temporaries to scalar temporaries, as support for vectorized temporaries has not yet been implemented
1. I changed memory hoisting to allow hoisting local memory outside of a mapGlobal, but the implementation will not work in the general case, its just a quick hack
1. if we merge this PR, we would end up with 3 different Harris applications in the codebase, which is a bit messy. things could be factorized.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ modules.xml
 *.pdf
 *.gz
 *.sc
+
+*.o
--- a/src/main/scala/apps/harrisCornerDetection.scala
+++ b/src/main/scala/apps/harrisCornerDetection.scala
@@ -12,6 +12,13 @@ import shine.OpenCL.KernelExecutor._
 
 import scala.reflect.ClassTag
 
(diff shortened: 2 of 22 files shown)


Bastacyclop:
Things to do before merging, matching the points listed above: 
1. run the code to check for correctness
2. could be merged separately
3. add support for temporary global memory
4. add support for vectorized temporaries?
5. open an issue for this and revert the hack
6. document why there are three different Harris versions

PR #231, opened 2022-04-27
Removed generic type projection use

michel-steuwer:
This is a new attempt to move towards Scala 3 by breaking preparation changes from #201 in multiple up-to-date PRs. With the hope that we can merge these quickly.

This PR removes the use of generic type projections (i.e. `T#U`) in the `FunctionHelper` class, as these have been [dropped](https://docs.scala-lang.org/scala3/reference/dropped-features/type-projection.html) in Scala 3.

The replacement syntax is maybe not ideal, but once in Scala 3 (...) there is a better alternative available: https://scastie.scala-lang.org/ndV4Xu8RQZWlZRYCc6GAig

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.4.7
+sbt.version=1.6.2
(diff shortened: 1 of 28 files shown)


Bastacyclop (approved):
Can we keep track of the desired Scala 3 alternative with an issue?

michel-steuwer:
Done: https://github.com/rise-lang/shine/issues/234

PR #232, opened 2022-04-27
Changed syntax as wildcard are invalid as backquoted identifiers in Scala 3

michel-steuwer:
Minor change to improve compatibility with Scala 3


Bastacyclop (commented):
I think I would reuse the sketch syntax here instead: `?n`, `?dt`, `?a`, ...

michel-steuwer:
> I think I would reuse the sketch syntax here instead: `?n`, `?dt`, `?a`, ...

Can you make that change and update the PR?

Bastacyclop (approved):

Bastacyclop:
@michel-steuwer done

PR #233, opened 2022-04-27
The type () is replaced with Unit

michel-steuwer:
The type () is replaced with Unit (as Scala 3 seems to no longer accept () as a type)

### DIFF
--- a/src/main/scala/rise/eqsat/Analysis.scala
+++ b/src/main/scala/rise/eqsat/Analysis.scala
@@ -12,8 +12,8 @@ trait Analysis {
   // FIXME: requiring non-type analysis is not supported
   def requiredAnalyses(): (Set[Analysis], Set[TypeAnalysis])
 
-  def init(egraph: EGraph): ()
-  def update(egraph: EGraph): ()
+  def init(egraph: EGraph): Unit
+  def update(egraph: EGraph): Unit
 
   // FIXME: we are currently not updating the analysis data to account for removals
   def eliminate(egraph: EGraph, toEliminate: HashSet[EClassId]): Unit = {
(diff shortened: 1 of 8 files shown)


src/main/scala/rise/eqsat/Analysis.scala
```diff
@@ -294,13 +294,13 @@ object NoAnalysis extends SemiLatticeAnalysis with TypeAnalysis {
   override def requiredTypeAnalyses(): Set[TypeAnalysis] = Set()
 
   override def make(egraph: EGraph, enode: ENode, t: TypeId,
-                    analysisOf: EClassId => ()): () = ()
+                    analysisOf: EClassId => Unit): Unit = ()
```
Bastacyclop, line 297:
```scala
object NoAnalysis extends SemiLatticeAnalysis with TypeAnalysis {
  type Data = ()
  type NatData = ()
  type TypeData = ()
```
Should these become `Unit` as well??

Bastacyclop (approved):

src/main/scala/rise/eqsat/Analysis.scala
michel-steuwer: Yes, I guess so. Everywhere where in the type `()` is used to mean Unit.
It's hard to search semantically like this in the codebase.

Could you make this update in the PR?

Bastacyclop (approved):

src/main/scala/rise/eqsat/Analysis.scala
Bastacyclop: @michel-steuwer should be good to merge now

Bastacyclop:
The CI seems to randomly fail due to missed nat simplifications :/

src/main/scala/rise/eqsat/Analysis.scala
michel-steuwer: Great. Thanks!

PR #238, opened 2023-04-25
Avoid anonymous objects, change unapply, remove unicode method names

michel-steuwer:


Bastacyclop:
I think these changes were basically merged through #245 , which includes #201 , not sure if there's stuff left to reuse here. @michel-steuwer

